PR #41, opened 2015-01-09
Hotfix/directory restructure

cored:
This is an experiment and open for discussion; my intention is to extract as much business logic that is couple to some entities in the system for instance ```User``` model. I got into this path of thought due to an error that the current test suite was hiding from us because of the mocking. If we start to use more isolated classes that iterate with each other we are going to be able to test this type of edge without mocking and reduce the brittleness of the specs. 

This is the list I would like to experiment with in this branch for this project: 
- [x] Directory for putting business logic ```/app/connect```
- [x] Directory for putting generic code which can then be extracted to be use as open source library ```/lib/connect```
- [x] Always use the ```Connect``` namespace; advantage of this would be to reuse common names for value or service objects. 
- [x] Namespace for use cases classes example ```Connect::Users```
- [ ] Remove business logic from models and use services instead; just put persistence logic inside AR models
- [ ] Test in isolation the business logic do integration test for AR models and feature specs
- [x] Decorate objects for adding new behavior - example ```Connect::Users::UserWithFullName```
- [x] For isolation testing use ```spec_helper``` for integration testing use ```rails_helper```


app/connect/users/access_token_freshner.rb
```diff
@@ -0,0 +1,46 @@
+module Connect
+  module Users
+    class AccessTokenFreshner
```
macat: Freshener

app/models/session.rb
```diff
@@ -1,3 +1,5 @@
+require_relative '../connect/users/token_expiry'
```
macat: I think rails includes this file automagically.

macat:
I see 3 changes here:
- The token freshener class, which makes sense as it implements the logic for refreshing the token. It also nicely wraps the implementation details.

- I'm not a big fan of `UserWithFullName`, because it puts a very simple logic in another file, making the source code harder to follow. A user will always have full name.

- The third change is the most confusing: `Connect::Users::TokenExpiry.for(access_token_expiry)`. When I read this line, I couldn't figure out what's happening here. Is this creating a new object? The token expiration time is closely coupled with the session, I'd keep it right there: `seconds.to_i.seconds.from_now` is much more obvious for other developers.

app/connect/users/token_expiry.rb
```diff
@@ -0,0 +1,9 @@
+module Connect 
+  module Users
+    class TokenExpiry
```
macat: @naush what do you think about this class?

app/connect/users/user_with_full_name.rb
```diff
@@ -0,0 +1,9 @@
+module Connect
+  module Users
+    class UserWithFullName < SimpleDelegator
+      def full_name
+        [first_name, last_name].compact.join(" ")
```
rwalters: The `first_name` and `last_name` are actually on the `Profile` model.

app/connect/users/token_expiry.rb
```diff
@@ -0,0 +1,9 @@
+module Connect 
+  module Users
```
rwalters: I noticed that there are `access_tokens`, `authorization_codes`, and `refresh_tokens` tables, each of which have a `token` field and an `expires_at` field. Should this be moved out of the `Users` namespace so that it could be used by other models that have tokens and expirations?

cored: @rwalters I don't follow the comment, those tables does not exist in Connect codebase.

app/connect/users/token_expiry.rb
naush: I see it is being used in more than one spot, which makes for a good argument for method extraction. However, I would make two suggestions. 1. The use case here seems generic enough that I feel like it can be extracted to a higher level, similar to ```app/lib/date_utility.rb```, where it can be reused throughout the rest of the code base without referencing ```Connect``` explicitly. 2. I would use time injection for this function so that we have more granular control over what is ```now``` (```Time.now``` or ```Time.current```, ie, do we care about timezone?), while also makes it easier to test.

cored:
Btw @naush @macat @rwalters this is my motivation for most of the changes in here: 
https://github.com/HouseTrip/guidelines/blob/master/sustainable-testing.md

Also I asked about one particular change regarding the decorator and the ```name``` method in here: 
https://github.com/HouseTrip/guidelines/issues/25

Hope this helps

app/connect/users/user_with_full_name.rb
macat: This is namely connect. There is no Profile object.

app/connect/users/token_expiry.rb
cored: @naush Will think about this changes; the reason why we need the namespace in this context is just for  making explicity that this is not Rails related code and it's more business logic one. Instead of ```app/lib``` I'm using ```/app/connect``` but the intention is the same; to me it's expose more that what that directory contains is business logic the ```/lib``` directory will contain generic code which can later be extracted to a gem if needed. 

I don't think we care about timezone at this stage and the particular use case was to remove coupling between the ```User``` model and the conversion logic; but I totally agree with you on injecting the ```Time```.

app/models/session.rb
naush: I am fine with explicit require if we decide to keep ```/connect``` out of autoloading path and we want to manage file loading order ourselves, but I would typically start from ```Rails.root``` instead of current path so that Rails doesn't accidentally load up the file twice if we require the same file in two different spots. They might have solved this problem already so then my concern would be irrelevant. But just in case...

cored: @naush My intention is to keep connect out of the autoloading path at first but was talking this over with @macat and maybe is not a good idea at this stage. My reasons to put it outside of that is because I again want to be as explicit as possible about external collaborators for each class.

app/connect/users/token_expiry.rb
naush: Given that all of our current plain ruby business domain objects live under ```app/lib``` right now, it does seem jarring to me to place ```connect/``` on the same level as ```models```, ```views```, and ```controllers```. I like the decision to pull these logic out of models and potentially move them to its own gem (which would make more sense for keeping this time utility function within the Connect code), but I would prefer it to live in ```app/lib/connect```, or in a gem, instead of ```app/```.

naush: Wait, I keep confusing this project with namely. Well, I guess you can make the decision of if you want to keep following the same pattern as we do or break out into your own thing with regarding to project structure. ;)

app/models/session.rb
naush: @cored I like explicit require, too! Clearer separation of concerns. You have to be totally commit to do it right though. Managing file loading order can be a headache!

naush:
@cored Thanks for sharing the conversation here! Given a established presentation layer, I would always choose to move my presentation logic there instead of models. They just don't belong in models!

I would add that, during the transition from a legacy Rails project, to one with service and presentation layers, the presentation code may be hiding in models, controllers, helpers, or views. In that case, I would keep following the existing pattern until I can pull out all the presentation code and consolidate them in a presentation object, one domain at a time. Just so the rest of the team won't be too confused when they try to find the new code.

Obviously this isn't an issue for Connect. I look forward to see how you implement presenters on this project!

app/models/session.rb
cored: @naush Agree, I'm planning on adding a coding guidelines for the project; but I have same concerns about you regarding manually loading dependencies for each classes; I don't know if I'm over complicating things but at some point I thought about a dependency injection framework for Ruby :-P

naush: Dangerous thoughts! :smirk_cat:

app/connect/users/token_expiry.rb
cored: @naush From my point of view ```app/connect``` suggest more business logic than ```app/lib``` because it's catch the attention of the developer faster ``lib`` can make some confusion with ```/lib```. Also trying to make things explicit in which we have the business logic by just looking at the directory structure. 

-> https://www.youtube.com/watch?v=WpkDN78P884

naush: @cored If we are starting from scratch here, I would actually put it under ```app/services``` or the like. Because to me ```app/**/``` designates the ```type``` of objects, for example: ```models```, ```views```, ```controllers```, ```mailers```, and not business domains or use cases. I agree that ```app/lib``` sounds more like where you would put utility functions in, but it's been around in the namely project for so long, it's become part of our tribal lingo.

cored: @naush Sure, that's why I said like creating a guideline to point this things out. ```app/connect``` will contain use cases like in this case would be what we normally put inside the ```app/services``` directory. My concern regarding using ```app/services``` and others directory is that this type of things doesn't create a relationship in the directory structure is just talking about related objects that doesn't have a physical relationship in disk. ```app/connect``` will contain use cases objects and all it's collaborators under a top level namespace which in this case will be ```Connect``` pointing out not just business logic but also the use case that we want to expose and reuse in some other places. In fact we even can extract the entire thing within a gem without a lot of trouble if we need to. Because the entire business logic will be in the same place. Uncle Bob's talk refer to this type of arch in Rails apps.

cored:
@naush My implementation is pretty straight forward; but I think what I'm using here are decorators I probably will introduce presenters when I get to the point in which i need to remove logic from the helpers in this app. I don't like helpers in Rails app neither because helpers represent a bunch of top level functions without any relationship whatsover. I agree with you in following the current patterns until moving the logic to the new scheme. In this project though, we can experiment a lot because is very small my intention is to have a save place to experiment and then take some patterns to use in others applications.

cored:
@naush @macat https://speakerdeck.com/skwp/domain-driven-rails

app/connect/users/access_token_freshner.rb
```diff
@@ -0,0 +1,46 @@
...
+      def self.fresh_access_token(user)
+        new(user).fresh_access_token
+      end
+
+      def initialize(user)
```
tabfugnic: Despite the fact that initialize is not directly a class method, consider moving this to the top of the stack

app/connect/users/access_token_freshner.rb
```diff
@@ -0,0 +1,46 @@
...
+        @user = user
+      end
+
+      def fresh_access_token
+        if access_token_expired?
+          refresh_access_token
+          user.save_token_info(refreshed_access_token,
```
tabfugnic: Consider changing the formate of this to

```ruby
user.save_token_info(
  refreshed_access_token,
  refreshed_expires_in
)
```

app/controllers/jobvite_connections_controller.rb
```diff
@@ -5,9 +7,14 @@ def edit
 
   def update
     @jobvite_connection = current_user.jobvite_connection
-    if @jobvite_connection.update(jobvite_connection_params)
+    connection_updater = Connect::Jobvite::ConnectionUpdater.
+      new(@jobvite_connection)
+
+    connection_updater.on(:connection_updated_successfully) do
```
tabfugnic: I don't fully understand why we need to use something like wisper. All we are doing is redirection and is very well suited the code that currently exists.

spec/connect/users/access_token_freshner_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    before do
+      stub_const("Authenticator", authenticator_class)
+    end
+
+    context 'when expire access token' do
+      let(:expires) { 1.day.ago }
+      let(:tokens) do 
+        { 'access_token' => 'refreshed_token',
```
tabfugnic: does this not work with symbols?

app/connect/users/access_token_freshner.rb
cored: @tabfugnic In terms of showing that the `fresh_access_token` is a factory method I think is better to have it in the top so people can see it as fast as possible, what would be the advantage to also put the initialize in the top or on top of that particular method ?

app/connect/users/access_token_freshner.rb
cored: @tabfugnic I think this type of changes to the code or ok; maybe we should try to define the styling using hound? Rubocop?

app/controllers/jobvite_connections_controller.rb
cored: @tabfugnic Here are the reasons for using something like Wisper: 

Extracting business logic from controllers; why wisper? I tried with different ways of implementing service objects within Rails applications all of them with advantages/disadvantages. The typical way of implementing a service object is like the following: 

```
class UserAuthenticator
  def initialize(user)
    @user = user
  end

  def authenticate(unencrypted_password)
    return false unless @user

    if BCrypt::Password.new(@user.password_digest) == unencrypted_password
      @user
    else
      false
    end
  end
end
```
Which is ok for testing in isolation which is what we want; but is breaks some principles that could bite us later on like "Tell, don't ask", because of the way the controller interacts with this piece of code. 

We can always use Matt Wayne [Hexagonal Rails](https://www.youtube.com/watch?v=CGN4RFkhH2M) way of doing service objects in which he includes two new methods in the controller definition and past the controller as a listener so the service call back the ```successful``` or ```unsuccessful```. The problem with this is that most controllers have more than one action in them that means that we are going to need two extra methods for each controller action. 

This leads me to using lambdas for passing the desire `redirect` or `render` behavior and calling those lambdas inside the service based on the conditional. Something like: 
```
class Controller < ApplicationController
   def create 
       Service.new(collaborator, params).execute(
         success: -> do |val| 
            # redirection logic
         end, 
         failure: -> do |val| 
           # render logic 
         end)
   end 
end
```

Which could also work; the downside for this is that I did not find a way to share behavior between the code that is running at that time; like logging or analytics. With wisper you can subscribe your service to other listeners with global code and it will work. Using wisper now will help us to have a mechanism for crossing concerns in the future which will be easily tested in isolation. 

Let me know what you think about this? Because my intention is to define an architecture for the app which I don't perceive we currently have.

tabfugnic: I understand what you are trying to do. I just am concerned that what is it does is make this more complicated to understand and feel isn't the best opportunity to show off how wisper will actually make this better.

cored: @tabfugnic Sorry Eric; I don't get your comment very well; you are saying that Wisper is actually going to make things better but at the same time it will be harder for understanding what it is doing? If so; what do you think we should try to aim for at this stage; because I really feel that mixing our business logic with framework code won't help us at the short/medium run taking into account that Connect is going to get a lot of changes in the near future. Let me know what are your thoughts on this and what you consider we can do about it.

tabfugnic: @cored Sorry, I didn't mean to confuse. I see this is as a premature optimization and only abstracting out some very straight forward code.

cored: @tabfugnic I get it; so you think the entire refactor is a premature optimization or just the implementation o service objects with wisper?

PR #50, opened 2015-02-11
Hotfix/connect error handling

cored:
This PR solve the way we display errors to the users after trying to import candidates from Jobvite.

### DIFF
--- a/app/presenters/jobvite/imports_presenter.rb
+++ b/app/presenters/jobvite/imports_presenter.rb
@@ -11,18 +11,18 @@ module Jobvite
     end
 
     def not_imported_candidates
-      import_results.to_a.map do |import_result| 
+      import_results.to_a.map do |import_result|
         format_error_message(import_result) if import_result[:result] =~ /error/i
       end.compact
     end
 
-    private 
+    private
+
+    attr_reader :import_results
 
     def format_error_message(import_result)
       import_result[:result] = import_result[:result].split(":")[-1]
       import_result
     end
-
-    attr_reader :import_results
   end
 end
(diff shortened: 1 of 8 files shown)


app/presenters/jobvite/imports_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    private 
+
+    def format_error_message(import_result)
+      import_result[:result] = import_result[:result].split(":")[-1]
+      import_result
+    end
+
+    attr_reader :import_results
```
rwalters: Could you put this at the top?

app/presenters/jobvite/imports_presenter.rb
```diff
@@ -0,0 +1,28 @@
+module Jobvite
+  class ImportsPresenter
+    def initialize(import_results)
+      @import_results = import_results
+    end
+
+    def imported_candidates
+      import_results.to_a.map do |import_result|
```
rwalters: You access this twice in the view.  Could you cache this in a variable?

I would also recommend you `select` what you want instead of having an `if` in the map followed by the `compact`. An example of both this recommendation and the caching:

```
def imported_candidates
  @imported_candidates ||= import_results.to_a.select do |result|
    import_result[:result] == I18n.t("status.success")
  end.map do |result|
    result[:candidate]
  end
end
```

app/presenters/jobvite/imports_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+        import_result[:candidate] if import_result[:result] == I18n.t("status.success")
+      end.compact
+    end
+
+    def not_imported_candidates
```
rwalters: Same recommendations as I made for `def imported_candidates`:

```
def not_imported_candidates
  @not_imported_candidates ||= import_results.to_a.select do |result|
    result[:result] =~ /error/i
  end.map do |result|
    format_error_message(result)
  end
end
```

macat:
+1

PR #53, opened 2015-03-11
Connect users to iCIMS account

tabfugnic:
* Implement an IcimsConnection model
* Users can connect iCIMS through the dashboard
* Clean up style violations throughout

### DIFF
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -2,27 +2,25 @@
 <section>
   <table>
     <tbody>
-      <tr>
+      <tr class="jobvite-account">
         <% if @jobvite_connection.connected? %>
           <td>
             <%= fa_icon "check-circle", class: "connected", text: "Jobvite" %>
             <p><%= @jobvite_connection.api_key %></p>
           </td>
-          <% if @jobvite_connection.missing_namely_field? %>
-            <td>
+          <td>
+            <% if @jobvite_connection.missing_namely_field? %>
               <%= t(".missing_namely_field", name: @jobvite_connection.required_namely_field) %>
-            </td>
-          <% else %>
-            <td></td>
-          <% end %>
+            <% end %>
+          </td>
           <td>
             <%= button_to t("dashboards.show.import_now"), jobvite_imports_path, class: '-n-btn' %>
           </td>
           <td>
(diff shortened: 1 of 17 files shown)


app/views/dashboards/show.html.erb
```diff
@@ -31,12 +29,22 @@
...
               <%= fa_icon "plug", text: t("dashboards.show.connect") %>
-            </a>
+            <% end %>
           </td>
         <% end %>
       </tr>
+
+      <tr class="icims-account">
+        <td>
+          <% unless @icims_connection.connected? %>
```
hrs: I favor `if !@icims_connection.connected?`.

hrs:
Most of this looks fine to me (though, having been involved in its construction, I'm biased).

My big concern is that there's a lot of duplication with the Jobvite code. Should we extract that somehow? Maybe a common superclass for the controllers or something?

app/views/dashboards/show.html.erb
tabfugnic: I just knew you would say that! My preference is `unless` but I can go either way. I assume the rest of the codebase probably wouldn't use `unless` so I will change it.

cored:
@hrs One of the reasons I started experimenting with this PR: [41](https://github.com/namely/connect/pull/41) was because of what you are saying in your comment; I feel that there is no architecture define for Connect. Can you take a look at that PR too; so I can get some feedback?

app/views/dashboards/show.html.erb
hrs: I guess an alternative would be `if @icims_connection.disconnected?`?

tabfugnic:
@hrs I see your concern and I think that as soon as this PR gets in we will start to be able to abstract some of this stuff out. I can do that now if you'd like, but I personally would like to wait as more things shake out.

@cored I will take a look.

hrs:
@tabfugnic Sure, I'm fine with giving things a little more time. Just gettin' the ball rolling on that conversation.

@cored I'll take a look, too!

cored:
@tabfugnic @hrs thanks

hrs:
Anyhoo, @tabfugnic, apart from that `unless`, I think this is good to merge.

PR #54, opened 2015-03-12
Users can disconnect their iCIMS account

hrs:
This just deletes their credentials. Jobvite disconnection works the same way.

### DIFF
--- a/app/models/icims/connection.rb
+++ b/app/models/icims/connection.rb
@@ -5,5 +5,12 @@ module Icims
     def connected?
       username.present? && password.present?
     end
+
+    def disconnect
+      update(
+        password: nil,
+        username: nil,
+      )
+    end
   end
 end
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -1,4 +1,5 @@
 <%= render "subheader" %>
+
 <section>
   <table>
     <tbody>
@@ -17,10 +18,7 @@
             <%= button_to t("dashboards.show.import_now"), jobvite_imports_path, class: '-n-btn' %>
           </td>
           <td>
-            <%= link_to jobvite_connection_path, class: "-n-btn -n-btn-sm -n-btn-danger -n-btn-link", method: :delete, rel: "nofollow" do %>
-              <%= fa_icon "times" %>
-              <%= t("dashboards.show.disconnect") %>
-            <% end %>
+            <%= render "disconnect_button", path: jobvite_connection_path %>
           </td>
         <% else %>
           <td>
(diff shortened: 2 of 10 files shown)


app/models/icims/connection.rb
```diff
@@ -5,5 +5,12 @@ class Connection < ActiveRecord::Base
     def connected?
       username.present? && password.present?
     end
+
+    def disconnect
+      update(
+        username: nil,
```
tabfugnic: :abc:

app/views/dashboards/show.html.erb
```diff
@@ -37,13 +38,28 @@
...
+          </td>
+          <td>
+            <%= link_to icims_connection_path, class: "-n-btn -n-btn-sm -n-btn-danger -n-btn-link", method: :delete, rel: "nofollow" do %>
+              <%= fa_icon "times" %>
+              <%= t("dashboards.show.disconnect") %>
+            <% end %>
+          </td>
+        <% else %>
+          <td>
+            <%= fa_icon "minus-circle", class: "disconnected", text: "Jobvite" %>
```
tabfugnic: `text: "Jobvite" probably want to make the Icims

app/views/dashboards/show.html.erb
```diff
@@ -37,13 +38,28 @@
...
       <tr class="icims-account">
-        <td>
-          <% if !@icims_connection.connected? %>
+        <% if @icims_connection.connected? %>
+          <td>
+            <%= fa_icon "check-circle", class: "connected", text: "iCIMS" %>
+            <p><%= @icims_connection.username %></p>
+          </td>
+          <td>
+            <%= link_to icims_connection_path, class: "-n-btn -n-btn-sm -n-btn-danger -n-btn-link", method: :delete, rel: "nofollow" do %>
```
tabfugnic: now that we've implemented this button twice can we turn this into a partial. Is it in our interested to actually turn it into a button? I don't really like hijacking `link_to` with `method: :delete` if memory serves me correctly it does that by hijacking the event with javascript

app/views/dashboards/show.html.erb
```diff
@@ -1,4 +1,5 @@
 <%= render "subheader" %>
+
```
tabfugnic: lol cramped

app/models/icims/connection.rb
hrs: Dammit, @tabfugnic, in *my* alphabet "username" comes before "password".

app/views/dashboards/show.html.erb
hrs: OOPS LOL

app/views/dashboards/show.html.erb
hrs: Yup, sounds good. `button_to` with `method: :delete` expands out to a form with a hidden field, right? Much better.

app/views/dashboards/show.html.erb
hrs: d421eb89c752cd0b5bee086983cdcf99efd7e01dc785cf0fe3e3951c7a88d440

app/views/dashboards/show.html.erb
```diff
@@ -17,10 +18,7 @@
             <%= button_to t("dashboards.show.import_now"), jobvite_imports_path, class: '-n-btn' %>
           </td>
           <td>
-            <%= link_to jobvite_connection_path, class: "-n-btn -n-btn-sm -n-btn-danger -n-btn-link", method: :delete, rel: "nofollow" do %>
-              <%= fa_icon "times" %>
-              <%= t("dashboards.show.disconnect") %>
-            <% end %>
+            <%= render "disconnect_button", path: jobvite_connection_path %>
```
tabfugnic: interesting I was thinking more in line with some sort of helper method, but this works too

tabfugnic:
keep-calm-and-you-may-fire-when-ready

app/views/dashboards/show.html.erb
hrs: Ah, interesting! I had thought *you'd* suggested rendering a partial.

PR #55, opened 2015-03-12
Institute rudimentary feature flagging

tabfugnic:


hrs:
Sure, I'm into this. My only concern would be having multiple features, in which case `SHOW_FEATURE` isn't the best name. Otherwise, :grapes:!

tabfugnic:
I figure we can cross that bridge when we come to it, but if you prefer I can rename it to SHOW_ICIMS

hrs:
Eh, no worries for now. Merge away!

PR #56, opened 2015-03-12
Change circle script to run migrations

tabfugnic:
* Circle doesn't run migrations on deploy
* We have explicitly run migration after deployment

https://circleci.com/docs/configuration#heroku-with-pre-or-post-deployment-steps

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,5 +5,6 @@ machine:
 deployment:
   production:
     branch: master
-    heroku:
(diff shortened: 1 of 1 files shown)


hrs:
Just to make sure, this is only executed if the tests pass, right? If so, it looks fine to me (though I'm not a CircleCI expert).

tabfugnic:
That's how the current circle ci thing works

PR #59, opened 2015-03-18
Check for icims field on namely

tabfugnic:
Check for field on namely employee that allows us to link back to the user from icims

### DIFF
--- a/app/models/icims/connection.rb
+++ b/app/models/icims/connection.rb
@@ -19,5 +19,36 @@ module Icims
         username: nil,
       )
     end
+
+    def missing_namely_field?
+      if connected?
+        check_namely_field
+      end
+    end
+
+    def required_namely_field
+      AttributeMapper.new.namely_identifier_field.to_s
+    end
+
+    private
+
+    delegate :namely_connection, to: :user
+
+    def check_namely_field
+      if namely_field_not_found? && namely_account_has_required_field?
+        update(found_namely_field: true)
+      end
+      namely_field_not_found?
+    end
+
+    def namely_field_not_found?
+      !found_namely_field?
+    end
+
+    def namely_account_has_required_field?
+      namely_connection.fields.all.detect do |field|
+        field.name == required_namely_field
+      end
+    end
   end
 end
--- a/spec/features/user_deletes_icims_connection_spec.rb
+++ b/spec/features/user_deletes_icims_connection_spec.rb
@@ -7,6 +7,7 @@ feature "User deletes iCIMS connection" do
       :icims_connection,
       :connected,
       user: user,
+      found_namely_field: true,
     )
 
     visit dashboard_path(as: user)
--- a/spec/features/user_visits_dashboard_spec.rb
+++ b/spec/features/user_visits_dashboard_spec.rb
@@ -7,13 +7,10 @@ feature "User visits their dashboard" do
       subdomain: ENV['TEST_NAMELY_SUBDOMAIN'],
     }
   end
-  let(:fixture_file) { "fields_without_jobvite" }
-  before do
-    stub_request(:get, /.*api\/v1\/profiles\/fields/)
-      .to_return(status: 200, body: File.read("spec/fixtures/api_responses/#{ fixture_file }.json"))
-  end
+
   context "with a Jobvite connection, but no Jobvite field on Namely" do
     scenario do
+      stub_namely_request("fields_without_jobvite")
       user = create(:user)
       create(
         :jobvite_connection,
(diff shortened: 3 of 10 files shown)


spec/models/icims/connection_spec.rb
```diff
@@ -52,4 +52,31 @@
       expect(icims_connection.password).to be_nil
     end
   end
+
+  describe "#missing_namely_field?" do
+    it "is doesn't check if not connected" do
```
hrs: Grammar!

spec/features/user_deletes_icims_connection_spec.rb
```diff
@@ -7,6 +7,7 @@
       :icims_connection,
       :connected,
       user: user,
+      found_namely_field: true
```
hrs: Trailing comma!

spec/models/icims/connection_spec.rb
```diff
@@ -52,4 +52,31 @@
...
+      icims_connection.missing_namely_field?
+
+      expect(icims_connection).to_not have_received(:check_namely_field)
+    end
+
+    it "checks and caches required namely field status" do
+      icims_connection = create(
+        :icims_connection,
+        :connected,
+        found_namely_field: false
```
hrs: Trailing comma!

hrs: Also, should we be assigning `found_namely_field` to something in the factory?

spec/models/icims/connection_spec.rb
```diff
@@ -52,4 +52,31 @@
...
   end
+
+  describe "#missing_namely_field?" do
+    it "is doesn't check if not connected" do
+      icims_connection = described_class.new
+      allow(icims_connection).to receive(:check_namely_field)
+
+      icims_connection.missing_namely_field?
+
+      expect(icims_connection).to_not have_received(:check_namely_field)
```
hrs: `not_to`?

app/models/icims/connection.rb
```diff
@@ -23,10 +23,35 @@ def disconnect
       )
     end
 
+    def missing_namely_field?
+      if connected?
+        check_namely_field
+        !found_namely_field?
```
elle: Would it read better if we renamed `found_namely_field?` to `namely_field_not_found?`?
It is used twice with the `!` oeprator

elle: A bit confusing that `found_namely_field?` appears here and inside the `check_namely_field` method

app/models/icims/connection.rb
```diff
@@ -23,10 +23,39 @@ def disconnect
...
+      end
+      namely_field_not_found?
+    end
+
+    def namely_field_not_found?
+      !found_namely_field?
+    end
+
+    def namely_account_has_required_field?
+      namely_connection.fields.all.detect do |field|
```
elle: Is `.all` needed here?

app/models/icims/connection.rb
```diff
@@ -23,10 +23,35 @@ def disconnect
...
     end
+
+    def check_namely_field
+      if !found_namely_field? && namely_account_has_required_field?
+        update(found_namely_field: true)
+      end
+    end
+
+    def namely_account_has_required_field?
+      namely_connection.fields.all.detect do |field|
```
elle: Actually, would this work?
`namely_connection.fields.any? do |field|`

app/models/icims/connection.rb
```diff
@@ -23,10 +23,35 @@ def disconnect
...
     private
 
+    delegate :namely_connection, to: :user
+
     def encoded_credentials
       Base64.encode64("#{username}:#{password}")
     end
+
+    def check_namely_field
+      if !found_namely_field? && namely_account_has_required_field?
```
elle: Wish that was simpler, but don't know enough to suggest something better

app/models/icims/connection.rb
tabfugnic: This is a part of the namely gem that we are using, all forces it into an array

spec/features/user_visits_dashboard_spec.rb
```diff
@@ -51,4 +48,50 @@
...
+
+  context "with a iCIMS connection, but no iCIMS field on Namely" do
+    scenario do
+      stub_namely_request("fields_without_icims")
+      user = create(:user)
+      create(
+        :icims_connection,
+        :connected,
+        user: user,
+        found_namely_field: false,
```
elle: Do we need to define it if it is `false` by default?

tabfugnic: despite it being the default, I'm putting it in there just so there is no ambiguity on what is being tested

elle:
Other than those comments, it looks good to me

PR #61, opened 2015-03-23
Require specific namely fields be returned

tabfugnic:
* Fields are email, firstname, lastname, gender and startdate

### DIFF
--- a/app/models/icims/candidate_find.rb
+++ b/app/models/icims/candidate_find.rb
@@ -1,0 +1,51 @@
+module Icims
+  class CandidateFind
+    REQUIRED_FIELDS = [
+      "email",
+      "firstname",
+      "gender",
+      "lastname",
+      "startdate",
+    ]
+
+    def initialize(connection:, authorized_request: AuthorizedRequest)
+      @connection = connection
+      @authorized_request = authorized_request
+    end
+
+    def find(id)
+      JSON.parse(get_candidate(id)).
+        merge("id" => id)
+    end
+
+    private
+
+    attr_reader :connection, :authorized_request
+
+    def get_candidate(person_id)
+      authorized_request.new(
+        connection: connection,
+        request: get_candidate_request(person_id)
+      ).execute
+    end
+
+    def get_candidate_request(person_id)
+      RestClient::Request.new(
+        method: :get,
+        url: person_url(person_id),
+        headers: {
+          key: connection.key,
+          params: { fields: required_person_fields },
+        },
+      )
+    end
+
+    def person_url(person_id)
+      "#{connection.api_url}/people/#{person_id}"
+    end
+
+    def required_person_fields
+      REQUIRED_FIELDS.join(",")
+    end
+  end
+end
--- a/app/models/icims/candidate_search.rb
+++ b/app/models/icims/candidate_search.rb
@@ -1,0 +1,45 @@
+module Icims
+  class CandidateSearch
+    HIREDATE_FIELD = "person.employeeinfo.hiredate"
+
+    def initialize(connection:, authorized_request: Icims::AuthorizedRequest)
+      @connection = connection
+      @authorized_request = authorized_request
+    end
+
+    def all
+      JSON.parse(search_people)
+    end
+
+    private
+
+    attr_reader :connection, :authorized_request
+
+    def search_people
+      authorized_request.new(
+        request: search_people_request,
+        connection: connection,
+      ).execute
+    end
+
+    def search_people_request
+      RestClient::Request.new(
+        method: :post,
+        url: "#{connection.api_url}/search/people",
+        data: search_params,
+        headers: { key: connection.key },
+      )
+    end
+
+    def search_params
+      {
+        filters: [
+          {
+            name: HIREDATE_FIELD,
+            operator: "=",
+          }
+        ]
+      }
+    end
+  end
+end
--- a/spec/models/icims/candidate_search_spec.rb
+++ b/spec/models/icims/candidate_search_spec.rb
@@ -1,0 +1,30 @@
+require "rails_helper"
+
+describe Icims::CandidateSearch do
+  describe "#all" do
+    it "get search results" do
+      stub_search_request
+
+      connection = double(
+        "icims_connection",
+        api_url: icims_customer_api_url,
+        key: "MY_KEY",
+        username: "USERNAME",
+      )
+
+      candidate_search = Icims::CandidateSearch.new(connection: connection)
+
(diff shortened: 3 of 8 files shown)


app/models/icims/candidate_find.rb
```diff
@@ -0,0 +1,54 @@
...
+      "#{connection.api_url}/people/#{person_id}"
+    end
+
+    def required_person_fields
+      [
+        "email",
+        "firstname",
+        "gender",
+        "lastname",
+        "startdate",
```
JoelQ: Thoughts on making this list a constant?

app/models/icims/candidate_search.rb
```diff
@@ -0,0 +1,43 @@
+module Icims
+  class CandidateSearch
+    def initialize(connection:, authorized_request: AuthorizedRequest)
```
JoelQ: What is an authorized request and why is it a class rather than an instance?

app/models/icims/candidate_search.rb
```diff
@@ -0,0 +1,43 @@
...
+    def all
+      JSON.parse(search_people)
+    end
+
+    private
+
+    attr_reader :connection, :authorized_request
+
+    def search_people
+      AuthorizedRequest.new(
```
JoelQ: Did you mean to hardcode `AuthorizedRequest` here?

app/models/icims/candidate_search.rb
```diff
@@ -0,0 +1,43 @@
...
+        data: search_params,
+        headers: { key: connection.key },
+      )
+    end
+
+    def search_params
+      {
+        filters: [
+          {
+            name: "person.employeeinfo.hiredate",
```
JoelQ: Thoughts on extracting this string to a named constant?

spec/models/icims/client_spec.rb
```diff
@@ -36,27 +40,27 @@
...
+            to_return(
+              body: '{"searchResults": []}',
+              headers: { "Content-Type" => "application/json" },
             )
+          connection = double(
+            "icims_connection",
+            key: "MY_KEY",
+            api_url: api_url,
+            username: "USERNAME",
+          )
```
croaky: Do we care about these options in this spec? Would `double("icims_connection").as_null_object` be enough to pass, and less likely to change later?

spec/models/icims/client_spec.rb
```diff
@@ -36,27 +40,27 @@
...
-              to_return(
-                body: '{"searchResults": []}',
-                headers: { "Content-Type" => "application/json" },
-              )
-            connection = double(
-              "icims_connection",
-              key: "MY_KEY",
-              api_url: api_url,
-              username: "USERNAME",
+    context "when the json response doesn't have candidates" do
```
croaky: "JSON"

spec/models/icims/candidate_search_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe "#all" do
+    it "get search results" do
+      stub_search_request
+
+      connection = double(
+        "icims_connection",
+        api_url: api_url,
+        key: "MY_KEY",
+        username: "USERNAME",
+      )
```
croaky: Same question about the `double` here and in spec/models/icims/candidate_find_spec.rb

spec/models/icims/candidate_search_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    {"searchResults" => [{ "id" => 8986 }]}.to_json
+  end
+
+  def hexdigest_matcher
+    /[a-f0-9]+/
+  end
+
+  def api_url
+    "https://api.icims.com/customers/2197"
+  end
```
croaky: Little bit of duplication with `hexdigest_matcher` and `api_url` across a couple of spec files. Dunno if it's worth extracting somewhere but one idea would be:

```ruby
# spec/support/icims_helpers.rb
module IcimsHelpers
  def hexdigest_matcher
    /[a-f0-9]+/
  end

  def icims_customer_api_url
    "https://api.icims.com/customers/2197"
  end
end

RSpec.configure do |config|
  config.include IcimsHelper, type: :model
end
```

app/models/icims/candidate_search.rb
tabfugnic: `AuthorizedRequest` is a decorator for a request that has yet to be built

app/models/icims/candidate_search.rb
tabfugnic: I did not! Woops

app/models/icims/candidate_find.rb
```diff
@@ -0,0 +1,54 @@
...
+      ].join(",")
+    end
+
```
croaky: Extra blank line? I thought Hound might catch these, but maybe not.

app/models/icims/candidate_search.rb
tabfugnic: Good plan

spec/models/icims/client_spec.rb
tabfugnic: I care about api_url, but not really anything else

croaky:
Looks good to me.

app/models/icims/candidate_find.rb
tabfugnic: I think hound got disabled at some point, not sure though

app/models/icims/candidate_find.rb
```diff
@@ -0,0 +1,54 @@
...
+          key: connection.key,
+          params: { fields: required_person_fields },
+        },
+      )
+    end
+
+    def candidate(person_id)
+      JSON.parse(get_candidate(person_id)).
+        merge("id" => person_id)
+    end
```
elle: Is `candidate` method really needed? i.e. call `JSON.parse...` from `find` method?

app/models/icims/candidate_find.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def find(id)
+      candidate(id)
+    end
+
+    private
+
+    attr_reader :connection, :authorized_request
+
+    def get_candidate(person_id)
```
elle: Thoughts on making `person_id` an `attr_reader`?

tabfugnic: I was on the fence about it, but I think I could do that

tabfugnic: On second thought, I think I will hold off for now.

spec/models/icims/client_spec.rb
elle: Just a question: why is `Timecop` needed here?

PR #62, opened 2015-03-23
Move to using HMAC only instead of Basic Auth

tabfugnic:


app/views/icims_connections/edit.html.erb
```diff
@@ -3,7 +3,7 @@
     <%= f.input :username, placeholder: "Username", input_html: { class: "half" } %>
   </div>
   <div class="form-row">
-    <%= f.input :password, placeholder: "Password", input_html: { class: "half" } %>
+    <%= f.input :key, placeholder: "Password", input_html: { class: "half" } %>
```
croaky: Should the placeholder also say "Key"?

spec/factories.rb
```diff
@@ -5,7 +5,7 @@
     trait :connected do
       customer_id 2187
       username "crashoverride"
```
croaky: Wow, "crashoverride". :clap:

croaky:
Looks simpler. Good to go.

PR #63, opened 2015-03-24
Use presenter for Icims create

tabfugnic:


app/presenters/imports_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+  def imported_candidates
+    @imported_candidates ||= import_results.to_a.select do |import_result|
+      import_result[:result] == I18n.t("status.success")
+    end.map do |import_result|
+      import_result[:candidate]
+    end
+  end
+
+  def not_imported_candidates
+    @not_imported_candidates ||= import_results.to_a.select do |import_result|
```
gabebw: Could this first `.select` block be extracted to a method?

app/presenters/imports_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+      format_error_message(import_result)
+    end
+  end
+
+  private
+
+  attr_reader :import_results
+
+  def format_error_message(import_result)
+    import_result[:result] = import_result[:result].split(":")[-1]
```
gabebw: What about `import_result.merge(:result, import_result[:result].split(":")[-1])`? Is that worse?

app/views/icims_imports/create.html.erb
```diff
@@ -1,29 +1,72 @@
...
+        <th><%= t(".contact_number_title") %></th>
+        <th><%= t(".email_address_title") %></th>
       </tr>
-    <% end %>
-  </tbody>
-</table>
+    </thead>
+    <tbody>
+      <% @icims_imports_presenter.imported_candidates.each do |candidate| %>
+        <tr>
```
gabebw: Is it worth making a partial for the candidate?

PR #64, opened 2015-03-24
Fix HMAC authentication

tabfugnic:
* Payload was not properly loading in request

### DIFF
--- a/app/models/icims/authorized_request.rb
+++ b/app/models/icims/authorized_request.rb
@@ -1,12 +1,13 @@
 module Icims
   class AuthorizedRequest < SimpleDelegator
-    alias request __getobj__
     HMAC_SHA256_HEADER = "x-icims-v1-hmac-sha256"
 
     def initialize(connection:, request:)
       super(request)
       @connection = connection
     end
+
+    alias :request :__getobj__
 
     def headers
       @headers ||= {
@@ -55,7 +56,7 @@ module Icims
         headers: headers.merge(
           "Authorization" => authorization_header,
         ),
-        data: payload,
+        payload: payload,
       ).execute
     end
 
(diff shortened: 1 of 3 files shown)


app/models/icims/authorized_request.rb
```diff
@@ -1,6 +1,7 @@
 module Icims
   class AuthorizedRequest < SimpleDelegator
     alias request __getobj__
```
JoelQ: Should this be `:request` ?

app/models/icims/authorized_request.rb
```diff
@@ -85,6 +86,10 @@ def canonical_query
       uri.query
     end
 
+    def payload
+      @payload ||= request.payload.to_s
```
JoelQ: Is it necessary to memoize here?

spec/models/icims/authorized_request_spec.rb
```diff
@@ -118,10 +118,10 @@ def canonical_request(request)
       #{request.method.upcase}
       /or-ya-know
       fields=stuff
-      content-type: application/json
```
gabebw: Why remove spaces?

app/models/icims/authorized_request.rb
tabfugnic: probably, but amazingly it hasn't complained at all

app/models/icims/authorized_request.rb
tabfugnic: Not really

spec/models/icims/authorized_request_spec.rb
tabfugnic: I should have explained this in my commit. They are doing an interesting form of api authentication that requires an exactly reproduceable hashing of about five different things. Unfortunately it's according to their spec that the spaces need to be removed. Otherwise I'd leave them in for readability.

app/models/icims/authorized_request.rb
tabfugnic: Oh no actually it is. RestClient does weird things with payload. It clears it after use or after another request. It's all very confusing. This I found worked the best

PR #65, opened 2015-03-24
Use contact phone that better reflects API

tabfugnic:


app/models/icims/candidate.rb
```diff
@@ -17,11 +17,20 @@ def name
     end
 
     def contact_number
-      phonenumber
+      if phones.present?
```
JoelQ: Can phones be `nil` ? If it is an empty array, can we get rid of the conditional?

tabfugnic: It can be nil

gabebw: What if we make a phones method:

```ruby
def phones
  @phones || []
end
```

Then I think we don't need this `if`?

tabfugnic: Sounds great to me

PR #66, opened 2015-03-25
User must have namely field to import candidates

tabfugnic:
* Change copy for missing namely field to be more strict
* Refactor connections on dashboard into partial
* Build presenters for connections to reduce duplication

### DIFF
--- a/app/presenters/icims/connection_presenter.rb
+++ b/app/presenters/icims/connection_presenter.rb
@@ -1,0 +1,25 @@
+module Icims
+  class ConnectionPresenter < SimpleDelegator
+    include ActiveModel::Conversion
+
+    def self.model_name
+      Connection.model_name
+    end
+
+    def route_key
+      self.class.model_name.singular_route_key
+    end
+
+    def readable_name
+      "iCIMS"
+    end
+
+    def namespace
+      :icims
+    end
+
+    def to_partial_path
+      "dashboards/connection"
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/presenters/icims/connection_presenter.rb
```diff
@@ -0,0 +1,29 @@
+module Icims
+  class ConnectionPresenter < SimpleDelegator
+    include ActiveModel::Conversion
+
+    def initialize(connection)
+      super(connection)
```
dgalarza: What's the reasoning behind having an initializer that just calls super?

app/presenters/jobvite/connection_presenter.rb
```diff
@@ -0,0 +1,29 @@
+module Jobvite
+  class ConnectionPresenter < SimpleDelegator
+    include ActiveModel::Conversion
+
+    def initialize(connection)
+      super(connection)
```
gabebw: I echo Damian's question above.

spec/features/user_visits_dashboard_spec.rb
```diff
@@ -22,10 +22,13 @@
 
       visit dashboard_path(as: user)
 
-      expect(page).to have_content t(
-        "dashboards.show.missing_namely_field",
-        name: "jobvite_id",
-      )
+      within(".jobvite-account") do
```
gabebw: Why doesn't this scenario have a descriptive name?

hrs: Good point. That was some other folks, but we'll fix that.

app/presenters/icims/connection_presenter.rb
hrs: I... don't know. I don't know.

PR #67, opened 2015-03-25
Gender should correspond with namely gender API

tabfugnic:


hrs:
I was thinking about extracting out some of this logic into a concern (or something), but I'm not really sure it's worthwhile.

Looks good to me!

PR #70, opened 2015-04-14
Hotfix/refactor icmis controller update

cored:


app/services/connection_updater.rb
```diff
@@ -0,0 +1,17 @@
+class ConnectionUpdater
```
rwalters: @cored I think this should keep a namespace of some sort. Otherwise, what connection is this updating? Socket connection of some sort? DB connection? `Jobvites` may not be correct, but perhaps `Api` or something?

cored: @rwalters Got it; yes. Probably you are right this is too generic.

cored:
@rwalters You know what, I think I would let it like this for now; because still need more info to take a decision I experiment with some names and did not feel right.

PR #74, opened 2015-04-15
Fix import error when no email exists

tabfugnic:


cored:
@tabfugnic Isn't this a case specifically for icims? Because as far as I know jobvite expose the missing fields in the non imported candidates table in the bottom of the page.

tabfugnic:
In this case it doesn't matter that much if it's for icims or for jobvite. This is just a presenter.

cored:
@tabfugnic I see; but still it doesn't feel that this presenter is been use by both imports because as I said when the import faileds on jobvite an error message saying that the import failed appears in the bottom table.

tabfugnic:
It is being used for both, on purpose. because it's a container for the candidate and the error message.

cored:
Cool +1

app/presenters/failed_import_candidate_presenter.rb
```diff
@@ -5,4 +5,8 @@ def initialize(candidate, error)
     super(candidate)
     @error = error
   end
+
+  def email
+    try(:email) || "noemail@example.com"
```
hrs: This smells like a NullObject.

hrs:
Cool, this looks good to me. I'm not sure if you want to implement the machinery for a NullObject just to clean up that one line. Merge at your discretion.

cored:
@hrs I don't feel at this stage that would be necessary the only thing that the null object will introduce will be another level of indirection because you are going just to move the nil check into another object. I think it's better to wait until we have more info on this particular problem.

PR #76, opened 2015-04-15
Add namely email to user

tabfugnic:
* script to back fill emails

### DIFF
--- a/lib/tasks/user_emails.rake
+++ b/lib/tasks/user_emails.rake
@@ -1,0 +1,11 @@
+namespace :user do
+  desc "Bring on emails for every user"
+  task emails: :environment do
+    User.all.each do |user|
+      profile = user.namely_connection.profiles.find(user.namely_user_id)
+      if user.email.nil? && profile.email
+        user.update(email: profile.email)
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


spec/models/session_spec.rb
```diff
@@ -36,6 +36,7 @@
         expect(user.access_token).to eq "my-access-token"
         expect(user.access_token_expiry).to be_within(1.second).of(expiry_time)
         expect(user.refresh_token).to eq "my-refresh-token"
+        expect(user.email).to eq "corgilover1965@example.com"
```
hrs: Whoa, there, this is Eugene Belford's email address. I'm pretty sure you want something like `"mr_the_plague@hotmail.com"`.

lib/tasks/user_emails.rake
```diff
@@ -0,0 +1,9 @@
+namespace :user do
+  desc "Bring on emails for every user"
+  task :emails => :environment do
```
hrs: New-style hash syntax?

lib/tasks/user_emails.rake
```diff
@@ -0,0 +1,9 @@
...
+    User.all.each do |user|
+      profile = user.namely_connection.profiles.find(user.namely_user_id)
```
hrs: I might add an `if profile.email` around this updating logic, just in case we ever need to run this again (like if something fails or times out for some reason).

hrs:
Aside from a few minor points, this looks fine to me.

PR #78, opened 2015-04-16
When adding to namely fails email user of failure

tabfugnic:
* Provide detailed error notes


hrs:
Looks straightforward to me. Merge away!

PR #79, opened 2015-04-16
Retry unsuccessful import and succeed

tabfugnic:
* Add link to email to retry
* Show view that retry was successful
* When import fails again error and explain reason


hrs:
Seems right to me! Merge away.

PR #80, opened 2015-04-17
Add salary and home_address

tabfugnic:


app/models/icims/candidate.rb
```diff
@@ -24,6 +24,29 @@ def to_partial_path
...
 
+    def home_address
+      if icims_home_address
+        {
+          address1: icims_home_address["addressstreet1"],
+          address2: icims_home_address["addressstreet2"],
+          city: icims_home_address["addresscity"],
+          country_id: icims_home_address["addresscountry"]["abbrev"],
+          state_id: icims_home_address["addressstate"]["abbrev"],
+          zip: icims_home_address["addresszip"],
```
hrs: Maybe introduce some short methods?

app/models/icims/candidate.rb
```diff
@@ -40,5 +63,17 @@ def phones
...
+
+    def icims_salary
+      self[:salary]
+    end
+
+    def icims_home_address
+      if self[:addresses]
+        @icims_home_address ||=
+          addresses.detect { |address|
+          address["addresstype"]["value"] == "Home"}
```
hrs: I might make this a little method, too. The formatting's a little off, too.

spec/models/icims/attribute_mapper_spec.rb
```diff
@@ -12,6 +12,8 @@
         lastname: "Murphy",
         start_date: "2014-01-02",
         gender: "Opt Out",
+        salary: "lots of money",
```
hrs: `"$$$TEXAS"`

app/models/icims/candidate.rb
```diff
@@ -24,6 +24,29 @@ def to_partial_path
       "icims_imports/candidate"
     end
 
+    def home_address
+      if icims_home_address
```
hrs: I notice there's no `else` associated with these conditionals. Is that a problem?

app/models/icims/attribute_mapper.rb
```diff
@@ -8,6 +8,8 @@ def call(icims_candidate)
         user_status: "active",
         start_date: icims_candidate.start_date,
         gender: namely_gender(icims_candidate.gender),
+        salary: icims_candidate.salary,
+        home: icims_candidate.home_address,
```
hrs: I don't know if you wanna factor out a service object for this one? It seems like there's a lot of address-mapping logic, and it might be nice to abstract that.

hrs: Only a suggestion, though. Feel free to ignore.

app/models/icims/candidate.rb
tabfugnic: I don't think so, we are currently mapping this to nil and if something is nil then it is going to be ignored by the mapper.

app/models/icims/candidate.rb
tabfugnic: Good point

PR #81, opened 2015-04-17
Add salary and home_address

tabfugnic:


tabfugnic:
This is just all sorts of wrong

PR #82, opened 2015-04-17
Change import now to use webhook instead

tabfugnic:
* Webhook url will take place of import now button
* Import now tests have been removed


spec/models/icims/connection_spec.rb
```diff
@@ -2,7 +2,10 @@
 
 describe Icims::Connection do
   describe "associations" do
+    subject { build(:icims_connection) }
     it { should belong_to(:user) }
+    it { should validate_presence_of(:user_id) }
+    it { should validate_uniqueness_of(:api_key) }
```
hrs: `expect` syntax?

config/routes.rb
```diff
@@ -8,7 +8,10 @@
   resources :jobvite_imports, only: [:create]
   resources :icims_imports, only: [:create]
 
-  resources :icims_candidate_imports, only: [:create]
+  post(
+    "/icims_candidate_imports/:api_key" => "icims_candidate_imports#create",
+    as: :icims_candidate_imports
```
hrs: TRAILING COMMA POLICE :police_car: :cop: :oncoming_police_car:

app/models/icims/connection.rb
```diff
@@ -30,6 +31,10 @@ def required_namely_field
       AttributeMapper.new.namely_identifier_field.to_s
     end
 
+    def set_api_key
+      self.api_key = SecureRandom.hex(10)
```
hrs: So when you create a connection, you generate a new API key. This is the key that the iCIMS webhook is expected to use to connect, right?

hrs:
Couple bits of minutia, but otherwise you're lookin' fine.

app/models/icims/connection.rb
tabfugnic: That is correct. Thoughts?

hrs: Just making sure I understood. I might make it 30 characters or so, because there's no reason not to.

PR #83, opened 2015-04-20
Fix items to work with icims

tabfugnic:
* remove CSRF token auth specifically for this end point
* Require baseline authentication by api token
* Fix attribute mapper to send over string instead of number

### DIFF
--- a/app/controllers/icims_candidate_imports_controller.rb
+++ b/app/controllers/icims_candidate_imports_controller.rb
@@ -1,5 +1,6 @@
 class IcimsCandidateImportsController < ApplicationController
   skip_before_filter :require_login
+  skip_before_filter :verify_authenticity_token
 
   def create
     candidate = Icims::Client.new(connection: connection).candidate(person_id)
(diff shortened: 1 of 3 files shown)


app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -1,5 +1,6 @@
 class IcimsCandidateImportsController < ApplicationController
   skip_before_filter :require_login
+  skip_before_filter :verify_authenticity_token
```
hrs: We can skip this step because we're looking up a connection by the api_key, right? So if the key doesn't exist, the request would fail anyway?

hrs:
You've got some failing tests, but otherwise this looks fine to me.

app/controllers/icims_candidate_imports_controller.rb
tabfugnic: That is correct. Rather than validating authenticity token, which I can't do cause it's a remote request. I am instead providing a token that must be verified and verified against the users icims customer id

PR #84, opened 2015-04-20
Unauthorized icims request

tabfugnic:
* Remove delay because it will always evaluate to be true

### DIFF
--- a/app/models/namely_importer.rb
+++ b/app/models/namely_importer.rb
@@ -34,7 +34,7 @@ class NamelyImporter
   def try_importing(attrs)
     if valid_attributes?(attrs)
       begin
-        namely_profiles.delay.create!(attrs)
+        namely_profiles.create!(attrs)
         SuccessfulCandidateImport.new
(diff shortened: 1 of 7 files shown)


app/views/icims_candidate_import_mailer/unauthorized_import.en.text.erb
```diff
@@ -0,0 +1,6 @@
+Icims has reported that you are unauthorized to access this account.
```
hrs: Don't they prefer "iCIMS"?

app/models/namely_importer.rb
```diff
@@ -34,7 +34,7 @@ def single_import(candidate)
   def try_importing(attrs)
     if valid_attributes?(attrs)
       begin
-        namely_profiles.delay.create!(attrs)
+        namely_profiles.create!(attrs)
```
hrs: Is this fast enough to be acceptable?

hrs:
Couple comments, but otherwise this looks good to go.

PR #85, opened 2015-04-23
Create migration to upgrade icims connections

tabfugnic:


macat:
This is still not going to fix the issue. If 2 users at the same time want to open the `/icims_connection/edit` page, the second user will get an error.

macat:
OK, nevermind. The connection model generates a new api key before creation.

PR #90, opened 2015-04-30
Hotfix/refactoring icmis imports

cored:


app/services/icims/candidate_importer.rb
```diff
@@ -0,0 +1,50 @@
+module Icims 
+  class CandidateImporter
+    attr_reader :candidate, :imported_result, :params
+
+    class FailedImport < StandardError; end
+    Params = Struct.new(:params) do
```
macat: What's the benefit of nesting parameters instead of setting it on `CandidateImporter`? Like `CandidateImporter#customer_id`

cored: Had mix feelings, but you are right.

app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -3,49 +3,20 @@ class IcimsCandidateImportsController < ApplicationController
...
-    import = namely_importer.single_import(candidate)
-
-    if import.success?
-      mailer.delay.successful_import(user, candidate)
-    else
-      mailer.delay.unsuccessful_import(user, candidate, import)
-    end
+    service = Icims::CandidateImporter.new(Icims::Connection, params)
+    service.import
+    mailer.delay.successful_import(service.user, service.candidate)
```
rwalters: @cored Why have the mailer in the controller? You pass in `service.user` and `service.candidate` here, and `service.imported_result` on an unsuccessful import, and `service.user` after rescuing an `Icims::Client::Error`. That indicates to me that the mailer should probably be called from the service. Would it work to move these Mailer calls into `Icims::CandidateImporter`?

If you want to keep the mailer out of `CandidateImporter`, I would recommend not `raise`-ing `FailedImport`, and instead checking on the import result status here to decide which mailer to send. It'd be one less error class floating around.

cored: @rwalters That was my first implementation; I left the mailer in the controller just because it's feels like part of Rails; but I guess I can just inject it and remove the communication with the controller for the importing logic because that's related to the service.

app/services/icims/candidate_importer.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def namely_importer
+      NamelyImporter.new(
+        namely_connection: user.namely_connection,
+        attribute_mapper: Icims::AttributeMapper.new,
+      )
+    end
+
+    def mailer
```
rwalters: @cored why not put mailer up in the `attr_reader` list?

rwalters:
@cored One question, but looks great otherwise!

+1

app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -3,49 +3,9 @@ class IcimsCandidateImportsController < ApplicationController
...
-    else
-      mailer.delay.unsuccessful_import(user, candidate, import)
-    end
-  rescue Icims::Client::Error => e
-    mailer.delay.unauthorized_import(user, e.message)
-  ensure
+    Icims::CandidateImporter.new(Icims::Connection, 
+                                 IcimsCandidateImportMailer, 
+                                 params).import
     render text: nil
```
viola: render nothing: true

app/services/icims/candidate_importer.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def api_key
+      params[:api_key]
+    end
+
+    def user
+      connection.user
+    end
+
+    def connection
```
viola: would it make sense to memoize this method, since it's used in two methods but wont change after initializing the object?

spec/services/icims/candidate_importer_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        allow_any_instance_of(Icims::Client).to receive(:candidate) { candidate }
+        allow(namely_profiles).
+          to(receive(:create!) { raise Namely::FailedRequestError.new })
+
+        expect(delayed).to receive(:unsuccessful_import)
+        service.import
+      end
+    end
+
+    context 'when unauthorized credentials in icims' do
```
viola: :+1: nice spec!

app/services/icims/candidate_importer.rb
cored: @viola good catch; it would hit twice the database otherwise.

app/services/icims/candidate_importer.rb
cored: @rwalters I just don't like to expose data that is not going to be use outside of the context of the object.

spec/services/icims/candidate_importer_spec.rb
cored: @viola thanks :-)

viola:
+1

PR #93, opened 2015-05-11
Hotfix/icims retry url no implicit conversion on personid

cored:


naush:
:+1:

app/presenters/icims/candidate_retry_imports_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+        imported_result.error
+      else
+        DEFAULT_ERROR_MESSAGE
+      end
+    end
+
+    private
+
+    attr_reader :imported_result, :candidate
+  end
```
viola: looks like both :imported_result, :candidate can be nil? how so - sorry just dont understand implemention

jasonfellows:
:+1:

app/presenters/icims/candidate_retry_imports_presenter.rb
cored: @viola yes, they can be nil. But I'm not sure if this is the correct implementation I feel that I'm hiding the actual error which is what I'm investing now after your comment. Thanks ;-)

app/presenters/icims/candidate_retry_imports_presenter.rb
```diff
@@ -0,0 +1,38 @@
+module Icims
+  class CandidateRetryImportsPresenter
+    DEFAULT_ERROR_MESSAGE = 'Check your email for details on errors'
+
+    def initialize(candidate, imported_result)
+      @candidate = candidate
```
jasonfellows: Would it be better to validate init params here before memory is allocated? Are there cases where we should be able to create an instance of this class when a param is nil?

app/presenters/icims/candidate_retry_imports_presenter.rb
viola: I see, sure thing @cored

app/presenters/icims/candidate_retry_imports_presenter.rb
cored: Yes probably, I feel the probably I got ahead of myself with the presenter.

rwalters:
+1

PR #94, opened 2015-06-08
Fix JavaScript tests

jferris:
The JavaScript tests use capybara-webkit, which requires a running X11
server. This adds xvfb to the Docker container and uses the Headless gem
to boot up an xvfb server as needed.


macat:
+1

cored:
+1

PR #96, opened 2015-06-11
Re-raise errors during tests

jferris:
This is the default behavior in Rails.

With show_exceptions set to true, integration tests will render a 500
page, hiding the exception and backtrace when an error occurs in a test.

With show_exceptions set to false, integration tests will immediately
fail when an error is raised, making it easier to debug and catch
errors.


base10:
LGTM

PR #97, opened 2015-06-15
Users can connect to NetSuite

jferris:
This commit implements a basic NetSuite client and follows a similar
workflow to other connections for setting up a NetSuite account.

The NetSuite connection is a little different, because it doesn't store
NetSuite credentials. It makes a request to Cloud Elements with the
user's credentials, and then stores an authorization token.

No import or export functionality is added for NetSuite. This commit
only adds the UI and backend for establishing a connection. It will
serve as a starting point for future commits.

The view doesn't show the NetSuite connection UI unless a Cloud Elements
API is configured. This means that development/test environments can see
the NetSuite UI, but staging and production will not until we're ready.

https://trello.com/c/HeQUdynL

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -3,7 +3,8 @@ class DashboardsController < ApplicationController
     @connections = [
       Jobvite::ConnectionPresenter.new(current_user.jobvite_connection),
       Icims::ConnectionPresenter.new(current_user.icims_connection),
-      Greenhouse::ConnectionPresenter.new(current_user.greenhouse_connection)
+      Greenhouse::ConnectionPresenter.new(current_user.greenhouse_connection),
+      NetSuite::ConnectionPresenter.new(current_user.net_suite_connection)
     ]
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,6 +31,8 @@ en:
         update: "Connect"
       jobvite_connection:
         update: "Connect"
+      net_suite_connection:
+        update: "Connect"
       namely_authentication:
         submit: "Sign in"
     link_to:
@@ -85,6 +87,14 @@ en:
   jobvite_connections:
     edit:
       title: "Jobvite"
+      slogan: "Please fill out the following fields to connect to this application."
+
+  net_suite_connections:
+    description:
+      connected: Connected NetSuite Description
+      disconnected: Disconnected NetSuite Description
+    edit:
+      title: "NetSuite"
       slogan: "Please fill out the following fields to connect to this application."
 
   jobvite_imports:
--- a/spec/models/net_suite/client_spec.rb
+++ b/spec/models/net_suite/client_spec.rb
@@ -1,0 +1,71 @@
+require "rails_helper"
+
+describe NetSuite::Client do
+  describe "#create_instance" do
+    context "on HTTP success" do
+      it "returns successful data" do
+        instance = { "id" => "123", "token" => "abcxyz" }
+        stub_request(
+          :post,
+          "https://api.cloud-elements.com/elements/api-v2/instances"
+        ).
+          with(
+            body: {
+              configuration: {
+                "user.username" => "u@example.com",
+                "user.password" => "secret",
+                "netsuite.accountId" => "123",
+                "netsuite.sandbox" => false
+              },
+              "element" => {
+                "key" => "netsuiteerp"
+              },
+              "tags" => [],
+              "name" => "123_netsuite"
+            }.to_json,
+            headers: {
+              "Authorization" => "User user-secret, Organization org-secret",
+              "Content-Type" => "application/json"
+            }
+          ).
+          to_return(status: 200, body: instance.to_json)
+
+        client = NetSuite::Client.new(
+          user_secret: "user-secret",
+          organization_secret: "org-secret"
+        )
+
+        result = client.create_instance(
+          email: "u@example.com",
+          password: "secret",
+          account_id: "123"
+        )
+
+        expect(result).to be_success
+        expect(result[:id]).to eq("123")
+        expect(result[:token]).to eq("abcxyz")
+      end
+    end
+
+    context "on HTTP failure" do
+      it "returns failure messages" do
+        error = "a failure"
+        stub_request(
+          :post,
+          "https://api.cloud-elements.com/elements/api-v2/instances"
+        ).
+          to_return(status: 400, body: { message: error }.to_json)
+
+        client = NetSuite::Client.new(
+          user_secret: "x",
+          organization_secret: "x"
+        )
+
+        result = client.create_instance({})
+
+        expect(result).not_to be_success
+        expect(result[:message]).to eq(error)
+      end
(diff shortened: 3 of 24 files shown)


app/controllers/dashboards_controller.rb
```diff
@@ -3,7 +3,8 @@ def show
     @connections = [
       Jobvite::ConnectionPresenter.new(current_user.jobvite_connection),
       Icims::ConnectionPresenter.new(current_user.icims_connection),
-      Greenhouse::ConnectionPresenter.new(current_user.greenhouse_connection)
+      Greenhouse::ConnectionPresenter.new(current_user.greenhouse_connection),
+      NetSuite::ConnectionPresenter.new(current_user.net_suite_connection)
```
base10: Is style on this project to use a trailing comma?

jferris: There wasn't a trailing comma before, so I didn't add one. It doesn't seem as though trailing commas are consistently used.

app/views/net_suite/connections/_connection.html.erb
```diff
@@ -0,0 +1,22 @@
+<% if ENV["CLOUD_ELEMENTS_ORGANIZATION_SECRET"].present? %>
```
base10: Perhaps use a helper here vs. exposing ENV in the view?

app/views/net_suite/connections/_connection.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <article class="connection net-suite-account">
+    <% if connection.connected? %>
+      <h3>
+        <%= fa_icon "check-circle", class: "connected", text: connection.readable_name %>
+      </h3>
+      <p>
+        NetSuite connected description
```
base10: I know the text here isn't final, but thoughts on moving this into i18n?

app/views/net_suite/connections/_connection.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      </p>
+    <% else %>
+      <%= link_to polymorphic_path([:edit, connection.route_key]), class: "-n-btn -n-btn-sm" do %>
+        <%= fa_icon "plug", text: t("dashboards.show.connect") %>
+      <% end %>
+      <h3>
+        <%= fa_icon "minus-circle", class: "disconnected", text: connection.readable_name%>
+      </h3>
+      <p>
+        NetSuite disconnected description
```
base10: I know the text here isn't final, but thoughts on moving this into i18n?

app/views/net_suite/connections/_connection.html.erb
```diff
@@ -0,0 +1,22 @@
+<% if ENV["CLOUD_ELEMENTS_ORGANIZATION_SECRET"].present? %>
+  <article class="connection net-suite-account">
+    <% if connection.connected? %>
```
base10: This seems like a good chunk of if/else logic in the view. Thoughts on further extracting these to partials and displaying the appropriate partial conditionally?

app/views/net_suite/connections/edit.html.erb
```diff
@@ -0,0 +1,27 @@
+<%= render "subnav" %>
+
+<section id="subheader">
+  <h1><%= t(".title") %></h2>
+  <p><%= t(".slogan") %></p>
+</section>
+
+<section>
+  <%= simple_form_for @net_suite_connection_request, url: net_suite_connection_path(@net_suite_connection), method: :put, html: {class: 'stacked'} do |f| %>
```
base10: Is house style to use the abbreviated `f` here or could we use `form` and avoid the abbreviation instead?

app/views/net_suite/connections/_connection.html.erb
jferris: I agree. I followed the general pattern established by other connectors ([example](https://github.com/namely/connect/blob/master/app/views/jobvite_connections/_connection.html.erb)), but I think the approach could use some refactoring. I also think there's some duplication between the various connectors.

jferris: That being said, I think that's likely a good follow-up pull request that would be good to make once we get the basic NetSuite export going.

app/views/net_suite/connections/_connection.html.erb
jferris: The view text for these partials hasn't been using I18n in general, but I'm not sure why.

@macat any thoughts here?

app/views/net_suite/connections/edit.html.erb
jferris: This markup follows the conventions used in other connection edit views.

config/locales/en.yml
```diff
@@ -110,6 +118,10 @@ en:
         api_key: "API key"
         hired_workflow_state: "Workflow state of hired candidates"
         secret: "Secret"
+      net_suite_connection:
+        email: "Email"
```
base10: Should these be in `config/locales/simple_form.en.yml` instead?

spec/models/net_suite/client_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+              },
+              "tags" => [],
+              "name" => "123_netsuite"
+            }.to_json,
+            headers: {
+              "Authorization" => "User user-secret, Organization org-secret",
+              "Content-Type" => "application/json"
+            }
+          ).
+          to_return(status: 200, body: instance.to_json)
```
base10: Add a space below to enhance readability?

spec/models/net_suite/client_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    context "on HTTP failure" do
+      it "returns failure messages" do
+        error = "a failure"
+        stub_request(
+          :post,
+          "https://api.cloud-elements.com/elements/api-v2/instances"
+        ).
+          to_return(status: 400, body: { message: error }.to_json)
```
base10: Add space for readability below?

config/locales/en.yml
jferris: There isn't a simple form locale file in this project yet. I'm seeing warnings when the app boots up, so I think configuration for simple form needs to be updated.

spec/models/net_suite/client_spec.rb
jferris: Done.

spec/models/net_suite/client_spec.rb
jferris: Done.

app/views/net_suite/connections/_connection.html.erb
jferris: I extracted this logic onto the `Connection` class in 55fd264.

base10: I like that a lot better.

app/views/net_suite/connections/_connection.html.erb
macat: I prefer using I18n if possible. Not sure why the other views didn't use it.

macat:
+1 This looks great!

PR #99, opened 2015-06-16
Use simple form's translations

base10:
* Initializes simple_form and a it's locale file
* Updates existing edit templates to use simple_form.en.yml
* For: https://trello.com/c/XR2HIFf2/29
* Tags: #rails

### DIFF
--- a/lib/templates/erb/scaffold/_form.html.erb
+++ b/lib/templates/erb/scaffold/_form.html.erb
@@ -1,0 +1,14 @@
+<%%= simple_form_for @<%= singular_table_name %>, html: { class: "stacked" } do |f| %>
+  <%- attributes.each do |attribute| -%>
+    <div class="form-row">
+      <%%= f.<%= attribute.reference? ? :association : :input %> :<%= attribute.name %>, input-html: { class: "half" } %>
+    </div>
+  <%- end -%>
+  </div>
+
+  <div class="-n-actions">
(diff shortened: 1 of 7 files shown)


lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
+  <%%= f.error_notification %>
+
+  <div class="form-inputs">
```
jferris: This looks different than the markup the existing forms actually use.

jferris:
Nice! Good to get those placeholders out of the markup.

lib/templates/erb/scaffold/_form.html.erb
base10: This is what simple form generates. You're suggesting we change this to match current style?

jferris: Right. That way, if somebody generates a new form, it will match the markup this project uses. Otherwise, somebody will need to modify the newly-generated form.

base10: Sounds good.

base10: Addressed with d540335. Buttons will use the plug and "Connect" text unless specifically overridden.

jferris:
This looks good to me.

PR #101, opened 2015-06-18
Refactors Jobvite and Greenhouse Connection controllers

base10:
* Adopts the pattern built with iCIMS and NetSuite connection
  controllers
* Greenhouse and Jobvite both have ::ConnectionForm objects
* Greenhouse::ConnectionForm is a little special in that a secret key is
  required, but is handled by the Greenhouse::Connection object that's
  passed in
  * secret_key is a read only value and is not edited through the form
    at all, therefore, it's not part of the validations on
    Greenhouse::ConnectionForm
* For: https://trello.com/c/VEnA88Nf/30


base10:
I think at this point, it possible to remove `ConnectionUpdater`. Open to doing that in this PR or in a separate one. Thoughts welcome.

jferris:
If `ConnectionUpdater` isn't being used anymore, I think it makes sense to remove here.

This looks good to me.

PR #102, opened 2015-06-18
Refactors submit helpers into the simple_form translations file

base10:
For: https://trello.com/c/mJTEBTWg/31


jferris:
Is it normal to put these helpers in the simple form file? They aren't actually used by simple_form - they're used by the regular Rails submit buttons, which simple_form uses.

base10:
@jferris I'm not able to argue normal. In my last project, we kept them in simple_form and I assumed that was standard. If we don't care to do this, I can just close the PR unmerged.

jferris:
I think I'd prefer to keep the simple_form translations file restricted to things in the `simple_form` namespace.

base10:
Fair enough.

PR #103, opened 2015-06-18
Improve Docker configuration

jferris:
Specifying the environment in docker-compose.yml made it impossible to
override secret keys in local .env files. Also, specifying the RAILS_ENV
and DATABASE_URL caused the development server and test to run in the
same database, meaning that running tests would wipe development data.

This also switches from Webrick to Unicorn, which is much faster and
doesn't hang when the Docker container goes to sleep.


macat:
LGTM

PR #104, opened 2015-06-19
Refactor to a single ConnectionsController

base10:
* All connections are now handled through a singular connections
  controller
* The previously existing connections controllers have been removed
* Routing has been adjusted so all of the existing routes are still
  used
* I also took the opportunity to reformat the routes file for
  readability
* For: https://trello.com/c/VEnA88Nf/30
* Tags: #rails


app/controllers/connections_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def connection
+    @connection ||= current_user.send(connection_type)
+  end
+
+  def connection_form_class_mapping
+    {
+      "greenhouse_connection" => Greenhouse::ConnectionForm,
+      "icims_connection" => Icims::ConnectionForm,
```
tabfugnic: Any thoughts on moving this to a FactoryClass?

base10: Hrm, interesting idea. I'll experiment with it.

base10: @tabfugnic Done in b905832

app/controllers/connections_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    params.require(connection_type).permit(
+      connection_form.class.allowed_parameters
+    )
+  end
+
+  def request_path
+    request.env.fetch("PATH_INFO")
+  end
+
+  def request_type
```
jferris: I think it would be worth using non-resource routes to avoid matching on the URL in the controller.

base10: Would you add a hidden form parameter then or write the URL such that you could extract a key?

jferris: I would change the URL.

app/controllers/connections_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    @connection ||= current_user.send(connection_type)
+  end
+
+  def edit_template
+    request_type.pluralize + "/edit"
+  end
+
+  def form_params
+    params.require(connection_type).permit(
+      connection_form.class.allowed_parameters
```
jferris: Thoughts on making this an instance method so you don't need to chain through the class?

app/controllers/connections_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def connection_form
+    @connection_form ||= ConnectionFormFactory.create(
+      connection: connection,
+      request_type: request_type
+    )
+  end
+
+  def connection_type
+    request_type.to_s
```
jferris: Won't `request_type` already be a string?

app/controllers/connections_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  def request_type
+    matches = request_path.match request_type_matcher
+    matches[1]
+  end
+
+  def request_type_matcher
+    /^\/(.*?)(\/edit)*$/
+  end
+
+  class ConnectionFormFactory
```
jferris: I think this is worth extracting and testing separately.

app/controllers/connections_controller.rb
base10: I could. Almost feels more like it should be a constant on the class, though, yes?

app/controllers/connections_controller.rb
base10: Indeed it is. This method also just gets to go away.

app/controllers/connections_controller.rb
jferris: I think there's a semantic argument for using a constant, but when you're using polymorphism to get varied behavior from a generic form object, I think it makes the most sense not to rely on the object's class. Using an instance method will require the same amount of code and won't be harder to invoke, and it will leave a lot of doors open: using runtime state to determine parameters, dynamic methods, and compositional patterns like decorators.

base10:
@jferris NRR

app/models/greenhouse/connection_form.rb
```diff
@@ -6,6 +6,10 @@ class ConnectionForm
 
     validates :name, presence: true
 
+    def allowed_parameters
```
jferris: Seems a little weird that these `allowed_parameters` methods are defined above `initialize`.

base10: Oh, whoops. Artifact of having converted them from class methods. I'll fix that up.

base10: Fixed in 1af0104

spec/services/connection_form_factory_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      context "failing" do
+        it "fails with an unknown form_type" do
+          connection = create("greenhouse_connection", :connected)
+          expect do
+            ConnectionFormFactory.create(
+              connection: connection,
+              form_type: "nonexistant_connection"
+            )
+          end.to raise_exception
```
jferris: I'd recommend specifying an exception type here. Using just `raise_exception` might cover up unexpected runtime errors from `nil` arguments or other common issues.

base10: Good idea.

spec/services/connection_form_factory_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        form_types = %w(
+          greenhouse_connection
+          icims_connection
+          jobvite_connection
+          net_suite_connection
+        )
+
+        form_types.each do |form_type|
+          it "allows #{form_type}" do
+            allow_form_type(form_type)
```
jferris: This overlaps with the coverage for each form type below. What do you think of this organization?

``` ruby
describe "#create_instance" do
  context "greenhouse" do
    it "returns a form"
  end

  context "icims" do
    it "returns a form"
  end

  context "jobvite" do
    it "returns a form"
  end

  context "net_suite" do
    it "returns a form"
  end

  context "unknown types" do
    it "raises an error"
  end
end
```

spec/services/connection_form_factory_spec.rb
base10: Addressed with 0ae97cb

spec/services/connection_form_factory_spec.rb
base10: Addressed with 2b96b09, good call.

base10:
@jferris NRR, when you have a moment.

app/services/connection_form_factory.rb
```diff
@@ -0,0 +1,54 @@
+class ConnectionFormFactory
+  def self.create(connection:, form_type:)
+    @connection = connection
```
jferris: I just noticed that these are all class methods. Since you're assigning instance variables in a class method, you're actually assigning instance variables for the class and not an instance. This means that these values persist between factory invocations. I found this somewhat confusing, and I think it could lead to bugs.

What do you think about:

* Making `create` just call `new(...).create_instance`
* Making the remaining methods instance methods instead of class methods

base10: Oh, sure. Good catch.

base10: Addressed in e653360

jferris:
Looks good!

PR #105, opened 2015-06-22
Greenhouse: custom field support [WIP]

macat:
This PR allows users to import profiles with custom fields from Greenhouse.


app/models/greenhouse/attribute_mapper.rb
```diff
@@ -1,6 +1,24 @@
...
+        .select { |_, value| value.present? }
+    end
+
+    def namely_identifier_field
+      :greenhouse_id
+    end
+
+    private
+
+    def basic_fields(payload)
```
base10: One of the things Joe mentioned to me on a separate project is making sure concepts within a method are at the same level of abstraction. Hound is calling out a high method complexity, and I agree. Basically, if you look at line 22, there's several methods to wade through before we see that we're working with a first_name.

First, I recommend setting payload as an instance variable once it hits call so it doesn't have to be passed all the way down. Next, I'd suggest turning call like `candidate_for(application_for(payload)).fetch("first_name")` into something like `candidate_field("first_name")` with "candidate_field() acting as a wrapper for the methods you're calling here.

You can carry that general concept through this method and improve the readability significantly.

app/models/greenhouse/attribute_mapper.rb
```diff
@@ -34,6 +51,20 @@ def application_for(payload)
       payload.fetch("application")
     end
 
+    def salary_field(payload)
+      offer = offer_for(payload)
+      custom_fields = offer.fetch("custom_fields", {})
+      if custom_fields.present? && salary = custom_fields.fetch("salary")
```
base10: I think you're assigning salary to custom_fields.fetch("salary") vs. testing for equality here. If that's what you want, I recommend moving the assignment above, similar to how you're setting `custom_fields`

app/models/greenhouse/attribute_mapper.rb
```diff
@@ -34,6 +51,20 @@ def application_for(payload)
...
+        {
```
base10: I might also move the value setting into its own method

app/services/greenhouse/candidates_importer.rb
```diff
@@ -23,8 +23,7 @@ def import
         if import.success?
           mailer.delay.successful_import(
             user,
-            candidate_name.to_s,
-            identified_custom_fields)
+            candidate_name.to_s)
```
base10: Style-wise, that closing paren is easier to see if it's on the next line and lines up with `mailer` on line 24

app/services/greenhouse/candidates_importer.rb
```diff
@@ -49,14 +48,11 @@ def user
...
-    def identified_custom_fields
-      CustomFieldsIdentifier.new(greenhouse_payload).field_names
-    end
-
     def namely_importer
       NamelyImporter.new(
         namely_connection: user.namely_connection,
-        attribute_mapper: Greenhouse::AttributeMapper.new
+        attribute_mapper: Greenhouse::AttributeMapper.new(
+          user.namely_connection.fields.all)
```
base10: See comment about closing param placement above.

app/services/greenhouse/custom_fields.rb
```diff
@@ -0,0 +1,31 @@
+module Greenhouse
+  class CustomFields
+    SUPPORTED_TYPES = ["single_select", "short_text", "long_text", "number"]
+
+    def self.match(payload, namely_fields)
+      new(payload, namely_fields).to_h
+    end
+
+    def initialize(payload, namely_fields)
```
base10: This is a great spot for using keyword arguments. We try to keep the keyword arguments in alphabetical order, but they're named and not positionally important.

spec/models/greenhouse/attribute_mapper_spec.rb
```diff
@@ -13,20 +22,17 @@
...
-        seasons: ["Season 1", "Season 2"]
+        salary: { yearly_amount: 80000, currency_type: "USD", date: "2015-01-23"},
       )
     end
 
     context "handle missing none mandatory fields" do
       it "return default values when not present in payload" do
-        greenhouse_candidate = JSON.parse(File.read("spec/fixtures/api_requests/greenhouse_payload_missing.json"))["payload"]
+        greenhouse_candidate = JSON.parse(
+          File.read("spec/fixtures/api_requests/greenhouse_payload_missing.json")
```
base10: @jferris added some helper methods here that would make sense to use here: `spec/support/features/namely_fields_helper.rb`

spec/requests/greenhouse_statuses_spec.rb
```diff
@@ -11,18 +11,28 @@
     create(:greenhouse_connection, :with_namely_field, name: "myhook") 
   end
 
+  before do
+    stub_request(:get, /#{ api_host }\/api\/v1\/profiles\/fields/).
+      to_return(
+        status: 200,
+        body: File.read("spec/fixtures/api_responses/fields_with_greenhouse.json")
```
base10: I recommend having a look at what Joe did with `spec/features/user_visits_dashboard_spec.rb` for a case like this.

spec/models/greenhouse/attribute_mapper_spec.rb
base10: When #112 get merged in, this will become `spec/support/features/namely_requests_helper.rb`

spec/services/greenhouse/custom_fields_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+    it "filters out not supported fields" do
+      expect(described_class.match(notsupported, namely_fields)).to eql({})
+    end
+    let(:payload_with_fields) do
+      {
+        "name_middle" => { "value" => "test",
+                           "type" => "short_text",
+                           "name" => "Middle name" },
+        "level" => { "value" => "ok",
```
base10: I recommend adding another line to separate lines 27 and 28.

spec/services/greenhouse/custom_fields_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+
+  describe "::match" do
+    let(:notsupported) do
+      {
+        "notsupported" => { "value" => "", "type" => "notsupported" },
+      }
+    end
+    it "filters out not supported fields" do
```
base10: Wrap this test with new lines

spec/services/greenhouse/custom_fields_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      expect(described_class.match(notsupported, namely_fields)).to eql({})
+    end
+    let(:payload_with_fields) do
```
base10: Any value in having the let blocks on either side of the test vs. together ahead of the test.

app/models/greenhouse/attribute_mapper.rb
macat: I improved the readability somewhat in #111 . This is a very weird structure and I didn't have time to refactor the whole class.

app/models/greenhouse/attribute_mapper.rb
macat: good point

spec/services/greenhouse/custom_fields_spec.rb
macat: moved the second let up.

PR #107, opened 2015-06-22
Extract duplication from Namely field check

jferris:
Each connection implements its own check for Namely fields, which
requires copying and pasting boilerplate code into each new integration.

This extracts the common code and uses the existing decorator in the
view so that new drivers only need to define the required field.

Also extends test coverage for the UserCheckNamelyField class.

### DIFF
--- a/spec/services/user_check_namely_field_spec.rb
+++ b/spec/services/user_check_namely_field_spec.rb
@@ -1,24 +1,93 @@
-require_relative '../../app/services/user_check_namely_field'
+require "spec_helper"
+require_relative "../../app/services/user_check_namely_field"
 
 describe UserCheckNamelyField do
-  subject(:user_check_namely_field) { described_class.new(connection) }
-  let(:connection) { double(
-    :connection,
-    found_namely_field?: found,
-    user: user,
-    required_namely_field: :field_name) }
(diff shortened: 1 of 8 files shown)


spec/services/user_check_namely_field_spec.rb
```diff
@@ -1,24 +1,93 @@
...
+  def stub_user(fields)
+    double(:user, namely_connection: stub_namely_connection(fields))
+  end
+
+  def stub_namely_connection(fields)
+    double(:namely_connection, fields: stub_fields_named(fields))
+  end
+
+  def stub_fields_named(names)
+    fields = names.map do |name|
```
jferris: Fixed in 78e3fc4.

spec/services/user_check_namely_field_spec.rb
```diff
@@ -1,24 +1,93 @@
-require_relative '../../app/services/user_check_namely_field'
+require "spec_helper"
```
base10: Any reason to prefer "spec_helper" vs. "rails_helper" here?

jferris: `spec_helper` has fewer dependencies and loads faster, so it's nice to use it when possible.

base10:
:sunglasses: Very nice. LGTM

PR #108, opened 2015-06-22
Require a NetSuite ID field on Namely

jferris:
This follows the same pattern as other integrations for determining
which Namely profile corresponds to which NetSuite employee.

This is required to prevent creating several employee records for a
single Namely profile.

Also refactors test fixture handling to consolidate logic for fetching
fields.

https://trello.com/c/wYAS4uMN

### DIFF
--- a/spec/support/features/namely_fields_helper.rb
+++ b/spec/support/features/namely_fields_helper.rb
@@ -1,0 +1,10 @@
+module Features
+  def stub_namely_fields(fixture_file)
+    stub_request(:get, %r{.*api/v1/profiles/fields}).
+      to_return(status: 200, body: namely_fields_fixture(fixture_file))
+  end
+
+  def namely_fields_fixture(fixture_file)
+    File.read("spec/fixtures/api_responses/#{fixture_file}.json")
+  end
+end
--- a/app/models/net_suite/connection.rb
+++ b/app/models/net_suite/connection.rb
@@ -8,4 +8,8 @@ class NetSuite::Connection < ActiveRecord::Base
   def enabled?
     ENV["CLOUD_ELEMENTS_ORGANIZATION_SECRET"].present?
   end
+
+  def required_namely_field
+    :netsuite_id
+  end
 end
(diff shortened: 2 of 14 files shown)


db/migrate/20150622200851_add_found_namely_found_to_net_suite_connections.rb
```diff
@@ -0,0 +1,11 @@
+class AddFoundNamelyFoundToNetSuiteConnections < ActiveRecord::Migration
```
base10: AddFoundNamelyField?

jferris: I believe the current naming follows conventions from [Rails documentation](http://edgeguides.rubyonrails.org/active_record_migrations.html#creating-a-migration):

> ``` ruby
> class AddPartNumberToProducts < ActiveRecord::Migration
>   def change
>     add_column :products, :part_number, :string
>   end
> end
> ````

base10: But the field is `:found_namely_field`, not `:found_namely_found`, so I'd expect the class and file name to match though, yes?

base10: I'd expect the full classname to be `AddFoundNamelyFieldtoNetSuiteConnections`, since I left that truncated.

spec/support/features/namely_fields_helper.rb
```diff
@@ -0,0 +1,10 @@
+module Features
+  def stub_namely_fields(fixture_file)
```
base10: Nice refactoring here.

base10:
@jferris One comment outstanding, otherwise LGTM.

db/migrate/20150622200851_add_found_namely_found_to_net_suite_connections.rb
jferris: Ah, I see - I wrote "Found" twice instead of "Field" the second time. Fixed in 1d0c679.

PR #109, opened 2015-06-23
Support for greenhouse short text salary field

macat:


app/models/greenhouse/attribute_mapper.rb
```diff
@@ -55,11 +55,20 @@ def salary_field(payload)
       offer = offer_for(payload)
       custom_fields = offer.fetch("custom_fields", {})
       if custom_fields.present? && salary = custom_fields.fetch("salary")
-        {
-          yearly_amount: salary.fetch("value").fetch("amount"),
-          currency_type: salary.fetch("value").fetch("unit"),
-          date: offer.fetch("starts_at")
-        }
+        if salary.fetch("type") == "currency"
```
base10: Over on #105, I left a comment about potentially extracting this to another method. I highly recommend that now that the logic is getting more involved.

PR #110, opened 2015-06-23
Support personal + work emails in Greenhouse

macat:


app/models/greenhouse/attribute_mapper.rb
```diff
@@ -21,7 +21,8 @@ def basic_fields(payload)
       {
         first_name: candidate_for(application_for(payload)).fetch("first_name"),
         last_name: candidate_for(application_for(payload)).fetch("last_name"),
-        email: email_for(candidate_for(application_for(payload))),
+        email: email_for(candidate_for(application_for(payload)), "work"),
```
base10: I recommend extracting this to a helper method or two to help readability.

spec/models/greenhouse/attribute_mapper_spec.rb
```diff
@@ -19,7 +19,8 @@
       expect(mapper.call(greenhouse_candidate)).to eq(
         first_name: "Johnny",
         last_name: "Smith",
-        email: "personal@example.com",
+        personal_email: "personal@example.com",
+        email: "work@example.com",
```
base10: This seems a little confusing if you're referring to this just as `email` vs. `work_email` here but as `work` in `app/models/greenhouse/attribute_mapper`

PR #111, opened 2015-06-23
Support for custom fields in job + offer (Greenhouse)

macat:


app/models/greenhouse/attribute_mapper.rb
```diff
@@ -19,54 +21,61 @@ def namely_identifier_field
 
     def basic_fields(payload)
       {
-        first_name: candidate_for(application_for(payload)).fetch("first_name"),
-        last_name: candidate_for(application_for(payload)).fetch("last_name"),
-        email: email_for(candidate_for(application_for(payload)), "work"),
-        personal_email: email_for(candidate_for(application_for(payload)), "personal"),
+        first_name: candidate(payload).fetch("first_name"),
```
base10: A potential abstraction for you is to build a candidate object that you could call methods. Would allow you to get away from having to remember its a data structure constantly.

macat: While I don't agree 100% with the current structure, it would be an overkill to abstract more object out of this class. This class has one purpose: converting GH's data structure to namely fields.

PR #114, opened 2015-06-23
Fix detection of NetSuite field in Namely

jferris:
This field wasn't being detected because of a symbol/string mismatch.


base10:
LGTM

PR #116, opened 2015-06-23
Record NetSuite internalIDs on Namely profiles

jferris:
This allows us to correlate which profile corresponds to which employee.

https://trello.com/c/MYmNpTNo


base10:
LGTM

PR #117, opened 2015-06-23
Update existing employees during NetSuite export

jferris:
When exporting a profile that was previously exported, update the
corresponding employee record in NetSuite.

https://trello.com/c/MYmNpTNo


base10:
LGTM

PR #118, opened 2015-06-24
Export job title when creating employee records at NetSuite

base10:
* Moves map_gender function to NetSuite::Export
* NetSuite::Export now handles data preparation for NetSuite::Client
* For: https://trello.com/c/CNi0jfKd/

### DIFF
--- a/app/models/net_suite/export.rb
+++ b/app/models/net_suite/export.rb
@@ -22,6 +22,12 @@ module NetSuite
     end
 
     class Employee
+      GENDER_MAP = {
+        "Male" => "_male",
+        "Female" => "_female",
+        "Not specified" => "_omitted",
+      }
+
       def initialize(profile, net_suite:)
         @profile = profile
         @net_suite = net_suite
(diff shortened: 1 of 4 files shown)


app/models/net_suite/export.rb
```diff
@@ -62,11 +72,13 @@ def id
...
-          last_name: @profile.last_name,
+          firstName: @profile.first_name,
+          lastName: @profile.last_name,
           email: @profile.email,
-          gender: @profile.gender,
-          phone: @profile.home_phone
+          gender: NetSuite::Export.map_gender(@profile.gender),
+          phone: @profile.home_phone,
+          subsidiary: { internalId: 1 },
+          title: @profile.job_title[:title]
```
base10: Question: Better way of handling @profile.job_title here?

spec/models/net_suite/export_spec.rb
```diff
@@ -1,6 +1,20 @@
 require "rails_helper"
 
 describe NetSuite::Export do
+  describe ".gender_map" do
```
jferris: Since this is directly mapping values from a hash and not actually calculating anything, I think it would be fine to not test this directly, and just verify that one gender comes through correctly in the unit test for `perform`.

app/models/net_suite/export.rb
```diff
@@ -1,5 +1,15 @@
 module NetSuite
   class Export
+    GENDER_MAP = {
```
jferris: If you end up removing the unit test for `map_gender`, you can move this constant and `map_gender` into `Employee` as private, which prevents this concern from leaking.

jferris:
I had a couple minor comments. Good to merge at your discretion.

PR #119, opened 2015-06-24
Refactor client initialization

jferris:
* Encapsulate ENV variables in one place
* Remove direct Client references in controllers
* Makes it easier initialize clients in several places
* Makes it easier to consolidate controllers

### DIFF
--- a/app/controllers/net_suite_exports_controller.rb
+++ b/app/controllers/net_suite_exports_controller.rb
@@ -2,18 +2,8 @@ class NetSuiteExportsController < ApplicationController
   def create
     export = NetSuite::Export.new(
       namely_profiles: current_user.namely_connection.profiles.all,
-      net_suite: net_suite_client
+      net_suite: current_user.net_suite_connection.client
     )
     @results = export.perform
-  end
-
-  private
-
-  def net_suite_client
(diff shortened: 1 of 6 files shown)


app/controllers/net_suite_exports_controller.rb
```diff
@@ -2,18 +2,8 @@ class NetSuiteExportsController < ApplicationController
   def create
     export = NetSuite::Export.new(
       namely_profiles: current_user.namely_connection.profiles.all,
-      net_suite: net_suite_client
+      net_suite: current_user.net_suite_connection.client
```
base10: Thoughts on having `current_user.net_suite_client` that calls `net_suite_connection.client` and keeps LOD happy?

jferris: In this case, I'm more concerned about how this looks as we add more integrations. If we need several methods (x_connection, x_client) on `User` for each integration, it's already up to 8 methods.

At some point it may be worth creating infrastructure so that integration controllers get a direct reference to the appropriate connection, so they don't have to go through the current user.

base10: Gotcha.

base10:
LGTM

PR #120, opened 2015-06-24
Export to every connected NetSuite account

jferris:
Adds a rake task to run an export for every NetSuite connection. Can be
used by Heroku scheduler to regularly run exports and keep Namely in
sync with NetSuite.

https://trello.com/c/k0OIdkkF

### DIFF
--- a/app/models/net_suite/bulk_export.rb
+++ b/app/models/net_suite/bulk_export.rb
@@ -1,0 +1,16 @@
+module NetSuite
+  class BulkExport
+    def initialize(users)
+      @users = users
+    end
+
+    def export
+      @users.ready_to_sync_with(:net_suite).each do |user|
+        Export.new(
+          namely_profiles: user.namely_profiles.all,
+          net_suite: user.net_suite_connection.client
+        ).perform
+      end
+    end
+  end
+end
--- a/spec/models/net_suite/bulk_export_spec.rb
+++ b/spec/models/net_suite/bulk_export_spec.rb
@@ -1,0 +1,60 @@
+require "rails_helper"
+
+describe NetSuite::BulkExport do
+  include Features
+
+  describe "#export" do
+    context "with a fully connected user" do
+      it "creates an export for every user with a NetSuite connection" do
+        user = create(:user)
+        create(
+          :net_suite_connection,
+          :connected,
+          :with_namely_field,
+          user: user
+        )
+        stub_namely_data("/profiles", "profiles_with_net_suite_fields")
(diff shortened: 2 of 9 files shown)


app/models/net_suite/bulk_export.rb
```diff
@@ -0,0 +1,16 @@
+module NetSuite
+  class BulkExport
+    def initialize(users)
+      @users = users
+    end
+
+    def export
+      @users.ready_to_sync_with(:net_suite).each do |user|
+        Export.new(
+          namely_profiles: user.namely_connection.profiles.all,
```
base10: Not crazy about the reach through here, but I think we covered that in an earlier PR.

spec/models/net_suite/bulk_export_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        stub_request(:put, %r{.*api/v1/profiles/.*}).to_return(status: 200)
+        stub_request(:any, %r{.*/api-v2/hubs/erp/employees.*}).
+          to_return(status: 200, body: { "internalId" => "123" }.to_json)
+
+        export user
+
+        expect(WebMock).to have_requested(
+          :post,
+          "https://api.cloud-elements.com/elements/api-v2/hubs/erp/employees"
+        ).twice
```
base10: That fixture file has three profiles. I take it we're bypassing the one that already has a netsuite_id set?

base10:
Couple of comments, but merge at your discretion

spec/models/net_suite/bulk_export_spec.rb
jferris: Yeah, the client issues a PUT for that instead.

This assertion it just here to make sure the export did something, as the details of the requests are covered by `export_spec.rb`. I'd rather do `at_least_once`, but that isn't supported by the matcher.

app/models/net_suite/bulk_export.rb
jferris: There's only going to ever be one Namely connection, and we don't look for many things from Namely (just fields and profiles, I think), so we could delegate those through the user.

spec/models/net_suite/bulk_export_spec.rb
base10: OK, sounds good.

app/models/net_suite/bulk_export.rb
base10: I think I'll like how that'll look delegated.

jferris: Updated in 8054588.

base10:
LGTM

PR #121, opened 2015-06-24
Send email enumerating the results of syncing with NetSuite

base10:
* Currently controller only
* Will list a count and names of employees successfully synced
* Will list a count and names of employees unsuccessfully synced
* For: https://trello.com/c/vlAopDqM/

### DIFF
--- a/app/controllers/net_suite_exports_controller.rb
+++ b/app/controllers/net_suite_exports_controller.rb
@@ -5,5 +5,15 @@ class NetSuiteExportsController < ApplicationController
       net_suite: current_user.net_suite_connection.client
     )
     @results = export.perform
+    send_results_email(@results)
+  end
+
+  private
+
+  def send_results_email(results)
+    SyncMailer.net_suite_notification(
+      email: current_user.email,
+      results: results
+    ).deliver
   end
 end
--- a/app/mailers/greenhouse_candidate_import_mailer.rb
+++ b/app/mailers/greenhouse_candidate_import_mailer.rb
@@ -1,7 +1,13 @@
 class GreenhouseCandidateImportMailer < ApplicationMailer
   def successful_import(user, candidate_name)
     @user = user
-    mail to: @user.email, subject: t(".subject", name: candidate_name)
+    mail(
+      to: @user.email,
+      subject: t(
+        "greenhouse_candidate_import_mailer.successful_import.subject",
+        name: candidate_name
+      )
+    )
   end
 
   def unsuccessful_import(user, candidate_name, status)
@@ -9,6 +15,12 @@ class GreenhouseCandidateImportMailer < ApplicationMailer
     @candidate_name = candidate_name
     @status = status
 
-    mail to: @user.email, subject: t(".subject", name: candidate_name)
+    mail(
+      to: @user.email,
+      subject: t(
+        "greenhouse_candidate_import_mailer.unsuccessful_import.subject",
+        name: candidate_name
+      )
+    )
   end
 end
--- a/app/mailers/sync_mailer.rb
+++ b/app/mailers/sync_mailer.rb
@@ -1,0 +1,52 @@
+class SyncMailer < ApplicationMailer
+  def net_suite_notification(email:, results:)
+    @sync_results = SyncResults.new(results)
+
+    mail(
+      to: email,
+      subject: t(
+        "sync_mailer.net_suite_notification.subject",
+        employees: @sync_results.employees(
+          @sync_results.succeeded.count
+        )
+      )
+    )
+  end
+
+  class SyncResults
+    def initialize(results)
+      @results = results
+    end
+
+    def succeeded
+      @succeeded ||= @results.select(&:success?)
+    end
+
+    def succeeded_message
+      I18n.t(
+        "sync_mailer.net_suite_notification.succeeded",
+        employees: employees(succeeded.count)
+      )
+    end
+
+    def failed
+      @failed ||= @results.reject(&:success?)
+    end
+
+    def failed_message
+      I18n.t(
+        "sync_mailer.net_suite_notification.failed",
+        employees: employees(failed.count)
+      )
+    end
+
+    def employees(count)
+      I18n.t(
+        "sync_mailer.net_suite_notification.employees",
+        count: count
+      )
+    end
+  end
+
+  private_constant :SyncResults
+end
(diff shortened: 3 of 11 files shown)


spec/features/user_exports_to_net_suite_spec.rb
```diff
@@ -34,5 +34,24 @@
     expect(page).to have_content("Exported updates for employee: Tina Tech")
     expect(page).to have_content("Couldn't export: Mickey Moore")
     expect(page).to have_content("Bad Data")
+
+    using_session :notification_email do
+      open_email user.email
+      current_email.save_and_open
```
base10: Extra line, needs to be removed.

base10:
Will fix translation test failure momentarily.

app/mailers/sync_mailer.rb
```diff
@@ -0,0 +1,45 @@
+class SyncMailer < ApplicationMailer
+  helper_method :failed, :failed_message, :succeeded, :succeeded_message
```
jferris: I think I'd wrap these up in an object and assign it for the view, rather than implementing them here and exposing them directly to the view.

app/controllers/net_suite_exports_controller.rb
```diff
@@ -5,5 +5,15 @@ def create
       net_suite: current_user.net_suite_connection.client
     )
     @results = export.perform
+    send_results_email(@results)
+  end
+
+  private
+
+  def send_results_email(results)
+    SyncMailer.delay.net_suite_notification(
```
jferris: We're going to be moving the entire export process into a background job, so I wouldn't worry about moving the mailer.

spec/features/user_exports_to_net_suite_spec.rb
```diff
@@ -34,5 +34,24 @@
     expect(page).to have_content("Exported updates for employee: Tina Tech")
     expect(page).to have_content("Couldn't export: Mickey Moore")
     expect(page).to have_content("Bad Data")
+
+    using_session :notification_email do
```
jferris: Any reason to start a new session?

base10: Looks like no. I thought I did, but I suspect where I've done this previously, I needed a separate session because I was clicking links and validating second pages.

app/controllers/net_suite_exports_controller.rb
base10: I get that we want to move the export process, but I'm not sure what action you're asking for here. Take the `delay` out perhaps?

jferris: Yes, sorry - I'd move `delay` out, as this will just create another background job, and I don't think it's worth the added complexity.

base10: Fixed in ac26d0e

spec/features/user_exports_to_net_suite_spec.rb
base10: Fixed in ac26d0e

app/mailers/sync_mailer.rb
base10: Fixed in ac26d0e

app/controllers/net_suite_exports_controller.rb
base10: Actually, not fixed. If I remove the `delay`, the test doesn't can't catch the email

jferris: You'll need to call `.deliver_now` at the end if you don't call `delay`.

base10: I tried that and `.deliver_later` before I tried `delay` and neither works:

`NoMethodError: undefined method `deliver_now' for #<Mail::Message:0x007f990ae51028>`

Investigating that further.

base10: Oh. We're using Rails 4.1.6, not 4.2.2. `.deliver_now` doesn't exist yet.

jferris: Ah, so it's just `deliver` then.

base10:
@jferris Ready for rereview.

app/mailers/greenhouse_candidate_import_mailer.rb
```diff
@@ -1,7 +1,13 @@
 class GreenhouseCandidateImportMailer < ApplicationMailer
   def successful_import(user, candidate_name)
     @user = user
-    mail to: @user.email, subject: t(".subject", name: candidate_name)
```
jferris: I'm curious why the other mailers changed.

app/mailers/sync_mailer.rb
```diff
@@ -0,0 +1,52 @@
...
+        plural_employees: plural_employees(succeeded.count)
+      )
+    end
+
+    def failed
+      @failed ||= @results.reject(&:success?)
+    end
+
+    def failed_message
+      I18n.t(
```
jferris: Rails/i18n has some functionality built in for this.

You can call `t("some_string", count: x)` and it will look for either `some_string.one` or `some_string.other` depending on the value of `x`.

http://edgeguides.rubyonrails.org/i18n.html#pluralization

app/mailers/greenhouse_candidate_import_mailer.rb
base10: Fun story.

If you use translations in mailers, you have to tell i18n to look at the mailers for translation use, otherwise they report as "unused" As soon as you do this, if you've used `t(".subject")` in another mailer, i18n-missing yells, because it looks for `t('mailer_class.subject)` from `t(".subject")` instead of `t("mailer_class.method.subject")`.

app/mailers/sync_mailer.rb
base10: Oh, spiffy. I'll use that, then.

app/mailers/greenhouse_candidate_import_mailer.rb
jferris: Ah, okay. Sounds like a fun journey.

base10: Lessons learned from my previous client project.

app/mailers/sync_mailer.rb
base10: Addressed with 64105ef, thanks!

jferris:
Looks good!

PR #123, opened 2015-06-25
Provide a notification when a user disconnects an integration

base10:
* In place for Greenhouse, iCIMS, Jobvite and NetSuite
  * NetSuite doesn't have its disconnect functionality wired in yet,
    so the functionality there is not wrapped with a test
  * I'm not overly worried about that fact since the same functionality
    is well covered with the disconnect feature specs for the other
    three current integrations
* Subsequent integrations will need to be added to INTEGRATION_MAP
  in ConnectionsController
* For: https://trello.com/c/fZ9WQynV/

### DIFF
--- a/app/mailers/connection_mailer.rb
+++ b/app/mailers/connection_mailer.rb
@@ -1,0 +1,19 @@
+class ConnectionMailer < ApplicationMailer
+  def authentication_notification(connection_type:, email:)
+    @integration = map_connection_type_to_integration(connection_type)
+
+    mail(
+      to: email,
+      subject: t(
+        "connection_mailer.authentication_notification.subject",
+        integration: @integration
+      )
+    )
+  end
+
+  private
+
+  def map_connection_type_to_integration(connection_type)
+    t("connection_mailer.integration_mapping.#{connection_type}")
+  end
+end
--- a/app/models/jobvite/client.rb
+++ b/app/models/jobvite/client.rb
@@ -1,7 +1,7 @@
 module Jobvite
   class Client
-    class Error < StandardError
-    end
+    class Error < StandardError; end
+    class Unauthorized < StandardError; end
 
     def self.recent_hires(connection)
       new(connection).recent_hires
@@ -12,7 +12,7 @@ module Jobvite
     end
 
     def recent_hires
-      CandidatePage.all(connection).flat_map { |page| page.candidates }
+      CandidatePage.all(connection).flat_map(&:candidates)
     end
 
     private
@@ -20,8 +20,12 @@ module Jobvite
     attr_reader :connection
 
     class CandidatePage
+      delegate :user, to: :connection
+
+      attr_reader :response_code
+
       def self.all(connection)
-        page = new(connection, 1)
+        page = new(connection: connection, start: 1)
         Enumerator.new do |yielder|
           while page
             yielder.yield page
@@ -30,22 +34,22 @@ module Jobvite
         end
       end
 
-      def initialize(connection, start)
+      def initialize(connection:, start: 1)
         @connection = connection
         @start = start
       end
 
       def candidates
-        if json_response.has_key?("errors")
-          raise Error, json_response["errors"]["messages"].to_sentence
-        else
-          json_response.fetch("candidates", {}).map { |hash| Candidate.new(hash) }
-        end
+        process_errors
+        json_response.fetch("candidates", {}).map { |hash| Candidate.new(hash) }
       end
 
       def next
         unless last_page?
-          self.class.new(connection, next_page_start)
+          self.class.new(
+            connection: connection,
+            start: next_page_start,
+          )
         end
       end
 
@@ -53,8 +57,32 @@ module Jobvite
 
       attr_reader :connection, :start
 
+      def invalid_secret_key
+        json_response.select do |key, value|
+          key == "status" && value == "INVALID_KEY_SECRET"
+        end
+      end
+
       def json_response
         @json_response ||= JSON.parse(RestClient.get(url))
+      end
+
+      def process_errors
+        raise_on_errors_message
+        raise_on_authenticaton_error
+      end
+
+      def raise_on_errors_message
+        if json_response.has_key?("errors")
+          raise Error, json_response["errors"]["messages"].to_sentence
+        end
+      end
+
+      def raise_on_authenticaton_error
+        if invalid_secret_key.present?
+          user.send_connection_notification("jobvite")
+          raise Unauthorized, json_response["responseMessage"]
+        end
       end
 
       def url
--- a/app/models/net_suite/client.rb
+++ b/app/models/net_suite/client.rb
@@ -2,14 +2,21 @@ module NetSuite
   class Client
     BASE_URL = "https://api.cloud-elements.com/elements/api-v2"
 
-    def self.from_env
+    def self.from_env(user)
       new(
+        user: user,
         user_secret: ENV["CLOUD_ELEMENTS_USER_SECRET"],
         organization_secret: ENV["CLOUD_ELEMENTS_ORGANIZATION_SECRET"]
       )
     end
 
-    def initialize(user_secret:, organization_secret:, element_secret: nil)
+    def initialize(
+      user:,
+      user_secret:,
+      organization_secret:,
+      element_secret: nil
+    )
+      @user = user
       @user_secret = user_secret
       @organization_secret = organization_secret
       @element_secret = element_secret
@@ -17,6 +24,7 @@ module NetSuite
 
     def authorize(element_secret)
       self.class.new(
+        user: @user,
         user_secret: @user_secret,
         organization_secret: @organization_secret,
         element_secret: element_secret
@@ -90,6 +98,9 @@ module NetSuite
       Result.new(true, response)
     rescue RestClient::BadRequest => exception
       Result.new(false, exception.response)
+    rescue RestClient::Unauthorized => exception
+      @user.send_connection_notification("net_suite")
+      raise Unauthorized, exception.message
     end
 
     def url(path)
@@ -137,6 +148,8 @@ module NetSuite
         @json ||= JSON.parse(@response)
       end
     end
+
+    class Unauthorized < StandardError; end
 
     private_constant :Result
   end
--- a/spec/mailers/connection_mailer_spec.rb
+++ b/spec/mailers/connection_mailer_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+describe ConnectionMailer do
+  it "provides integration name in the subject, addresses the email, and " \
+  "describes the problem" do
+    connection_type = "icims"
+    email = "test@example.com"
+    mailer = ConnectionMailer.authentication_notification(
+      connection_type: connection_type,
+      email: email
+    )
+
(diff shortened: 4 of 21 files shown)


app/controllers/connections_controller.rb
```diff
@@ -1,4 +1,11 @@
 class ConnectionsController < ApplicationController
+  INTEGRATION_MAP = {
```
jferris: Instead of using a hash for this, what do you think about using an I18n lookup based on the key in the mailer view?

base10: Sure, I can do that.

base10: Addressed in 9b626cf. NB: I opted to reuse existing translation keys. Let me know if you think we need something purpose specific, even if it's duplicative.

app/models/jobvite/client.rb
```diff
@@ -30,31 +37,50 @@ def self.all(connection)
...
+      attr_reader :connection, :start, :user
 
       def json_response
-        @json_response ||= JSON.parse(RestClient.get(url))
+        @json_response ||= JSON.parse(
+          RestClient.get(url)
+        )
+      end
+
+      def process_errors
```
base10: Addressed with 46628da

app/models/jobvite/client.rb
```diff
@@ -1,27 +1,34 @@
...
 
-    def initialize(connection)
+    def initialize(connection:, user:)
       @connection = connection
+      @user = user
     end
 
     def recent_hires
-      CandidatePage.all(connection).flat_map { |page| page.candidates }
+      CandidatePage.all(
```
base10: Addressed with 46628da

app/models/jobvite/client.rb
base10: Addressed with 46628da

app/mailers/connection_mailer.rb
```diff
@@ -0,0 +1,19 @@
...
+        "connection_mailer.authentication_notification.subject",
+        integration: @integration
+      )
+    )
+  end
+
+  private
+
+  def map_connection_type_to_integration(connection_type)
+    t("#{connection_type}_connections.edit.title")
```
jferris: It may be better to make a new key instead of using this key, or to rename it to something more generic. I could see somebody changing `edit.title` to something like "Edit NetSuite Settings" without realizing it will also change the text in emails.

app/models/jobvite/bulk_import.rb
```diff
@@ -22,7 +22,7 @@ def new_importer(user)
       Importer.new(
         user,
         connection: user.jobvite_connection,
-        client: Client.new(user.jobvite_connection),
+        client: Client.new(connection: user.jobvite_connection, user: user),
```
jferris: Was the Jobvite client also modified? I'm having a hard time figuring out where these parameters would go: https://github.com/namely/connect/blob/nlw-notify-of-lost-sync-credentials/app/models/jobvite/client.rb

app/models/net_suite/client.rb
```diff
@@ -2,14 +2,21 @@ module NetSuite
   class Client
     BASE_URL = "https://api.cloud-elements.com/elements/api-v2"
 
-    def self.from_env
+    def self.from_env(user)
       new(
+        user: user,
```
jferris: This is only used by `authorize`. Thoughts on passing the user there instead of the initialize?

app/models/jobvite/bulk_import.rb
base10: Actually, I pulled that back out, so I need to fix this up.

spec/mailers/connection_mailer_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+describe ConnectionMailer do
+  it "provides integration name in the subject" do
+    connection_type = "icims"
```
jferris: The set up for these examples looks like it's the same. Thoughts on merging these all into one `it` block?

spec/models/icims/authorized_request_spec.rb
```diff
@@ -106,6 +130,21 @@ def connection
     @connection ||= build(:icims_connection, :connected)
   end
 
+  def user
+    @user ||= @connection.user
```
jferris: This looks like a reimplementation of `let`. I'd recommend either:

* Using `let` if we planning on going in that direction
* Restructuring this spec file to avoid instance state

spec/models/jobvite/client_spec.rb
```diff
@@ -119,15 +127,58 @@
...
     end
+
+    context "on authentication failure" do
+      it "returns invalid authentication message" do
+        stub_request(:get, "https://api.jobvite.com/api/v2/candidate").
+          with(
+            query: hash_including("api" => "MY_API_KEY", "sc" => "MY_SECRET")
+          ).
+          to_return(
+            body: '{
```
jferris: What do you think about using a heredoc here, doc? (har har har)

``` ruby
to_return(
  body: <<-JSON
{...}
  JSON
)
```

spec/services/greenhouse/candidates_importer_spec.rb
```diff
@@ -45,6 +46,22 @@
             candidates_importer.import
           }.to raise_error(Greenhouse::CandidatesImporter::Unauthorized)
         end
+
+        it "sends an invalid authentication message" do
+          allow_any_instance_of(Greenhouse::ValidRequesterPolicy).to(
```
jferris: It took me a few scans to figure out this invocation. I'm not sure if it's clearer or if I'm just used to it, but I think it's the parenthesis after `to` that's throwing me off. What do you think of line-splitting after the dot so you don't need parens?

``` ruby
allow_any_instance_of(Greenhouse::ValidRequesterPolicy).
  to receive(:valid?).and_return(false)
```

jferris:
Some high level ideas/questions:

* What do you think about putting a method on `User` to deliver notifications? There's some duplicated boilerplate around setting up the mailer that might be nice to delegate. Not meaningful duplication, but it would still be tidier.
* What happens to these exceptions after the email is raised? Will the user end up seeing a 500 page? How do they fix their credentials once they get messed up?

app/models/net_suite/client.rb
base10: Tried this out, but doesn't seem like a good fit, since ultimately I'm expecting to find `user` on that object when `wrap_response` gets called

app/models/jobvite/bulk_import.rb
base10: Fixed in e56df47

spec/mailers/connection_mailer_spec.rb
base10: Think I'd feel better about that since there's no way to get around having to trap the exception.

app/models/net_suite/client.rb
jferris: Ah, I see - the two methods looked like one in the diff view.

spec/mailers/connection_mailer_spec.rb
base10: Fixed with da1e471

spec/models/icims/authorized_request_spec.rb
base10: Fixed in `954b17b`, although, if you look at Line 129, we're doing the same sort of thing with connection, which I'm using here. That method is used in a test helper method. Shall I address that, too?

jferris: Yeah, I think it would be good to restructure to remove the instance state in general.

spec/models/jobvite/client_spec.rb
base10: Done in cff37a6

spec/services/greenhouse/candidates_importer_spec.rb
base10: Fixed with 9fc5db8

spec/models/icims/authorized_request_spec.rb
base10: OK, will do.

base10: Actually, spending a few minutes with unwinding instance state on this test, I agree it's worthwhile, but this seems like a larger task since there's instance state used in several places. I think refactoring this should be a separate PR.

Opened https://trello.com/c/58rvUSyp/ to address this.

app/mailers/connection_mailer.rb
base10: Good call. Fixed up in 2daeafd

base10:
@jferris On your [feedback](https://github.com/namely/connect/pull/123#issuecomment-117273718)

1. Good call, I was trying to think of how to centralize this better. I like this approach and will be implementing it.

2. I expect if the user is importing through the imports controller, they'll get a 500. Thinking it through, though, that's something we should rescue and give them instructions on addressing. We should also address this when the background task runs into bad authentication, but that's possibly out of scope here. Would like your thoughts on that.

The way I'd been thinking about a user addressing the issue by deleting the connection and reestablish it for now. We should quickly iterate on this to provide true edit functionality.

jferris:
@base10 Those `User` changes look good to me.

> I expect if the user is importing through the imports controller, they'll get a 500. Thinking it through, though, that's something we should rescue and give them instructions on addressing. ... The way I'd been thinking about a user addressing the issue by deleting the connection and reestablish it for now.

Yeah, I think that would be good to do as part of this story. Even if we just bring them back to the dashboard with a flash message, it would be useful to tell them "you need to disconnect and reconnect" and not just "I'm sorry, but something went wrong."

> We should also address this when the background task runs into bad authentication, but that's possibly out of scope here.

I think we should probably do that now, as the main reason for wanting emails is to know when an automated/background sync fails. However, I think it's fine to break out a new story for that; I just assume it's next on the priority list.

> We should quickly iterate on this to provide true edit functionality.

Yeah, I think it would be good to throw a story in there for that.

base10:
@jferris Will address controller handling. For the other pieces, I've opened up the following Trello cards:

* Edit connections: https://trello.com/c/xrYEiY3u
* Keep automated/background sync running: https://trello.com/c/I6f1verE

base10:
OK, so that was full of discovery.

Jobvite will now present a flash message and return the user to the dashboard in case of authentication error.

The other three controllers aren't fits for this workflow for the following reasons:

* iCIMS and Greenhouse don't import directly. Instead, they're reached by webhooks, so presenting a flash message serves no benefit and indeed, that workflow isn't reachable
  * In this case, I suspect we need a separate card to present an error status to the calling service
* NetSuite promptly enqueues to DelayedJob, so I suspect any exceptions raised by the NetSuite exporter won't be seen by the controller at all

jferris:
Alright, makes sense. After working through these a little more, it seems like the integrations can be divided into two categories:

* Push sync (on-demand, background)
* Pull sync (subscribe and receive notifications for updates)

For push sync, it can be further divided into import/export, but that shouldn't make much difference.

There may be further refactoring we can do in the controllers and service classes to make it easier to reuse the error handling and authentication management. I'll take a look at that shortly.

base10:
@jferris Any further review comments, then?

jferris:
Looks good!

base10:
\o/ Now, to rebase and merge this sucker.

PR #125, opened 2015-06-25
Users can disconnect NetSuite accounts

jferris:
Adds the standard "Disconnect" button for NetSuite connections.

https://trello.com/c/tyoiiyDQ


base10:
LGTM

PR #126, opened 2015-06-30
Prepare for connection view refactoring with some i18n cleanup

base10:
*  Move long strings to use "folded style"
  * More info:
    * http://symfony.com/doc/current/components/yaml/yaml_format.html
* Alphabetize the integrations
* For: https://trello.com/c/SkFozk1d

PR #127, opened 2015-06-30
Extract and standardize descriptions for each integration

base10:
* Each integration will be expected to have a connected_html and a
  disconnected_html description localization
* I'm using `_html` since some of the descriptions will have links and
we want ActionView to render said links
* For: https://trello.com/c/SkFozk1d
* Follow-up to https://github.com/namely/connect/pull/126
  * NB: The above case should be merged and rebased first since they
    have a common diff (cherry-picked a2a54a3 from that branch)


jferris:
This looks good. Let's just make sure to get the WIP text captured somewhere.

PR #128, opened 2015-06-30
Connections use new/create instead of edit/update

jferris:
* We're actually creating new connections in most cases
* We need the ability to actually edit an existing connection


base10:
LGTM. Tying in with what i'm doing on #123, I concur that we need the ability to edit existing connections.

PR #129, opened 2015-06-30
Require a subsidiary after connecting NetSuite

jferris:
Introduces an alternate path for connections where the edit action is
used until the connection has enough feels to be ready.

For NetSuite connections, presents a list of possible subsidiaries and
exports profiles as employees under the selected subsidiary.

https://trello.com/c/uaW9epLf

### DIFF
--- a/app/controllers/connections_controller.rb
+++ b/app/controllers/connections_controller.rb
@@ -6,9 +6,22 @@ class ConnectionsController < ApplicationController
 
   def create
     if connection_form.update(form_params)
-      redirect_to dashboard_path
+      redirect_to after_save_path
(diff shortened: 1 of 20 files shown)


base10:
Overall, code looks good to me. I wanted to confirm that we can't do this through the normal new/create cycle because we need the credentials set-up first in order to request the subsidiary list.

jferris:
> I wanted to confirm that we can't do this through the normal new/create cycle because we need the credentials set-up first in order to request the subsidiary list.

That's right. We can't see the subsidiaries for a NetSuite account until we have access to the account.

PR #132, opened 2015-07-01
Eliminate connection presenters

jferris:
These presenters were largely dead code, as the ActiveModel pieces were
either unused or provided by the ActiveRecord objects they decorated.

Two methods were used:

* readable_name, which is replaced by i18n
* namespace, which was moved directly to the connection objects

I also renamed "namespace" to "integration_id" for greater parity with
routes and Rails conventions.


config/locales/en.yml
```diff
@@ -159,6 +159,18 @@ en:
         We are exporting your profiles to NetSuite and will send you an email as
         soon as the export is complete.
 
+  greenhouse:
```
base10: You could also eliminate some duplication under connection_mailer.integration_mapping with this same approach.

jferris: I'm not finding the `integration_mapping` method. Is that merged yet?

base10: Sorry, that's at line 60 on `master`

jferris: Oh, I needed to rebase. Updated in a4132e0.

base10:
LGTM

PR #133, opened 2015-07-01
Make NetSuite sync more generic

jferris:
This extracts the NetSuite-specific parts of the export controller,
route, view, mailer, and background job.

We can move other on-demand integrations (Jobvite) into this framework
by adding a button to the right URL and exposing a `sync` method on the
`Connection` class.

https://trello.com/c/jU8MIB1L


base10:
Looks sharp!

PR #134, opened 2015-07-02
Log an error when an integration can't proceed due to auth issues

base10:
* With SyncJob, this traps the exception
* Icims::CandidatesImporter was trapping the exception, now it logs
* Greenhouse::CandidatesImporter does not trap the exception, which
  was the expected behavior previous to notifying the user
  * Greenhouse::CandidatesImporter will now log the error
* Jobvite does not yet have an implementation that would be affected by
  this
* For: https://trello.com/c/I6f1verE/

### DIFF
--- a/app/jobs/sync_job.rb
+++ b/app/jobs/sync_job.rb
@@ -6,7 +6,10 @@ class SyncJob
 
   def perform
     results = connection.sync
-    deliver_notification results
+    deliver_sync_notification results
+  rescue Unauthorized => exception
+    log_unauthorized_exception(exception)
+    deliver_unauthorized_notification
   end
 
   private
@@ -15,12 +18,27 @@ class SyncJob
     user.send(:"#{@integration_id}_connection")
   end
 
-  def deliver_notification(results)
+  def deliver_sync_notification(results)
     SyncMailer.sync_notification(
       email: user.email,
       integration_id: @integration_id,
       results: results
     ).deliver
+  end
+
+  def deliver_unauthorized_notification
+    user.send_connection_notification(@integration_id)
+  end
+
+  def integration_name
+    I18n.t("#{@integration_id}.name")
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for user_id: #{user.id} " \
+      "with #{integration_name}"
+    )
   end
 
   def user
--- a/app/models/net_suite/client.rb
+++ b/app/models/net_suite/client.rb
@@ -99,7 +99,6 @@ module NetSuite
     rescue RestClient::BadRequest => exception
       Result.new(false, exception.response)
     rescue RestClient::Unauthorized => exception
-      @user.send_connection_notification("net_suite")
       raise Unauthorized, exception.message
     end
 
@@ -148,8 +147,6 @@ module NetSuite
         @json ||= JSON.parse(@response)
       end
     end
-
-    class Unauthorized < StandardError; end
 
     private_constant :Result
   end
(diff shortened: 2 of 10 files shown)


base10:
Ideas welcome on some of the exception logging duplication.

app/jobs/sync_job.rb
```diff
@@ -16,14 +19,27 @@ def connection
   end
 
   def deliver_notification(results)
-    SyncMailer.sync_notification(
-      email: user.email,
-      integration_id: @integration_id,
-      results: results
-    ).deliver
+    if !erred?
```
jferris: I don't think the `erred?` logic is necessary, because an exception will skip the `deliver_notification` and jump into the `rescue` clause of `perform`.

app/jobs/sync_job.rb
```diff
@@ -2,11 +2,14 @@ class SyncJob
   def initialize(integration_id, user_id)
     @integration_id = integration_id
     @user_id = user_id
+    @erred = false
   end
 
   def perform
     results = connection.sync
     deliver_notification results
+  rescue NetSuite::Client::Unauthorized => exception
```
jferris: Since this is a generic class (not specific to NetSuite), do we want to make a generic `Unauthorized` exception and use it here?

app/jobs/sync_job.rb
```diff
@@ -16,14 +19,27 @@ def connection
...
+        results: results
+      ).deliver
+    end
+  end
+
+  def handle_authentication_exception(exception)
+    @erred = true
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for user_id: #{user.id}"
+    )
```
jferris: Do we want to deliver a notification to the user here?

app/jobs/sync_job.rb
base10: I was seeing something different when I was working on this as `NetSuite::Export::Job`, but let me give that a shot.

base10: Tested and works without `erred?`. Yay. Fixed with 278fb6a

app/jobs/sync_job.rb
base10: A user will get a message when the exception is raised in `NetSuite::Client`

app/jobs/sync_job.rb
base10: I did like having <Integration>::<class>::Unauthorized across the four exceptions since that made it clear where the exception was raised, but I'm open to consolidating them to one exception class.

app/jobs/sync_job.rb
jferris: If we handle it here instead of in the clients, we'll only have to implement it once for NetSuite/Jobvite, right? And future, on-demand integrations will get it "for free?"

base10: The expectation is that the import controller is going to effectively run a sync job vs. calling the import directly, I'm good with that.

base10: Addressed with 3fbf492

app/jobs/sync_job.rb
base10: Addressed with 9b5298d. Thoughts on consolidating the other integration exceptions in this PR vs. a separate one / as needed?

base10:
@jferris Feedback addressed, ready for rereview. Have one question for you: https://github.com/namely/connect/pull/134#discussion_r33792524

app/jobs/sync_job.rb
jferris: I think it would be fine to do in a separate pull request. I think it's a good idea, though - every integration will have to handle authorization errors, so we may as well have a single exception that represents the common concept.

jferris:
Ready to squash and merge at your discretion.

base10:
Thank you, @jferris!

PR #135, opened 2015-07-06
Extract common authentication notification methods to a module

base10:
* Convert Greenhouse::CandidatesImporter to use newly extracted
authentication methods
* Let SyncJob use the common methods as well
* This leaves iCIMS in place for this PR, since that API looks to be
more involved and worth handling in a focused PR
* There's likely a subsequent PR to further align Greenhouse and Icims
candidate importing
* For: https://trello.com/c/NGozb2N4/

### DIFF
--- a/app/jobs/sync_job.rb
+++ b/app/jobs/sync_job.rb
@@ -8,8 +8,7 @@ class SyncJob
     results = connection.sync
     deliver_sync_notification results
   rescue Unauthorized => exception
-    log_unauthorized_exception(exception)
-    deliver_unauthorized_notification
+    notifier.log_and_notify_of_unauthorized_exception(exception)
   end
 
   private
@@ -26,22 +25,16 @@ class SyncJob
     ).deliver
   end
 
-  def deliver_unauthorized_notification
-    user.send_connection_notification(@integration_id)
-  end
-
-  def integration_name
-    I18n.t("#{@integration_id}.name")
-  end
-
-  def log_unauthorized_exception(exception)
-    Rails.logger.error(
-      "#{exception.class} error #{exception.message} for user_id: #{user.id} " \
-      "with #{integration_name}"
+  def notifier
+    AuthenticationNotifier.new(
+      integration_id: integration_id,
+      user: user
     )
   end
 
   def user
     @user ||= User.find(@user_id)
   end
+
+  attr_reader :integration_id
 end
--- a/app/models/authentication_notifier.rb
+++ b/app/models/authentication_notifier.rb
@@ -1,0 +1,30 @@
+class AuthenticationNotifier
+  def initialize(integration_id:, user:)
+    @integration_id = integration_id
+    @user = user
+  end
+
+  def integration_name
+    I18n.t("#{integration_id}.name")
+  end
+
+  def log_and_notify_of_unauthorized_exception(exception)
+    log_unauthorized_exception(exception)
+    deliver_unauthorized_notification
+  end
+
+  private
+
+  def deliver_unauthorized_notification
+    user.send_connection_notification(integration_id)
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for " \
+      "user_id: #{user.id} with #{integration_name}"
+    )
+  end
+
+  attr_reader :integration_id, :user
+end
--- a/app/services/greenhouse/candidates_importer.rb
+++ b/app/services/greenhouse/candidates_importer.rb
@@ -1,8 +1,12 @@
 module Greenhouse
   class CandidatesImporter
-    attr_reader :mailer, :params, :signature, :secret_key, :connection_repo
-
-    class Unauthorized < StandardError; end
+    attr_reader(
+      :connection_repo,
+      :mailer,
+      :params,
+      :secret_key,
+      :signature,
+    )
 
     def initialize(mailer, connection_repo, params, secret_key, signature)
       @params = params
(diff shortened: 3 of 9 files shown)


app/models/integration_authentication_notifications.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for " \
+      "user_id: #{user.id} with #{integration_name}"
+    )
+  end
+
+  def set_exception
```
jferris: It looks like this method is only used in `CandidatesImporter`. Does it need to be in this module?

app/models/integration_authentication_notifications.rb
```diff
@@ -0,0 +1,25 @@
+module IntegrationAuthenticationNotifications
```
jferris: Thoughts on extracting a class here (composition) instead of mixing in a module (inheritance)?

It's hard for me to understand the interface between the module and client classes. I think a separate class would be clear, though, as I could understand shared state from the initializer and shared behavior from public methods.

app/models/integration_authentication_notifications.rb
base10: In this PR, that's true. I believe I moved it thinking this PR was going to be larger before deciding to take a more iterative approach.

In later PRs, I have it in `app/models/icims/authorized_request.rb` and `app/models/jobvite/client.rb`. 

I could move this out until those other PRs come in and move it over at the appropriate time.

app/models/integration_authentication_notifications.rb
base10: Sure. I'll look at doing that.

base10: @jferris Addressed with  97b6543. That's a lot clearer. Thanks for the feedback.

app/jobs/sync_job.rb
```diff
@@ -8,8 +8,8 @@ def perform
     results = connection.sync
     deliver_sync_notification results
   rescue Unauthorized => exception
-    log_unauthorized_exception(exception)
-    deliver_unauthorized_notification
+    notifier.log_unauthorized_exception(exception)
+    notifier.deliver_unauthorized_notification
```
jferris: Can we extract this pair of calls to a single public method on notifier? It doesn't look like we ever call one without immediately calling the other.

app/jobs/sync_job.rb
```diff
@@ -26,22 +26,16 @@ def deliver_sync_notification(results)
...
-  def integration_name
-    I18n.t("#{@integration_id}.name")
-  end
-
-  def log_unauthorized_exception(exception)
-    Rails.logger.error(
-      "#{exception.class} error #{exception.message} for user_id: #{user.id} " \
-      "with #{integration_name}"
+  def notifier
+    @notifier ||= AuthenticationNotifier.new(
```
jferris: Any reason to memoize?

app/jobs/sync_job.rb
base10: Sure can.

app/services/greenhouse/candidates_importer.rb
```diff
@@ -54,15 +59,18 @@ def greenhouse_payload
...
-      Rails.logger.error(
-        "#{exception_class} error #{error_message} for user_id: #{user.id}"
-      )
+      raise exception
+    end
 
-      raise Unauthorized, error_message
+    def notifier
+      @notifier ||= AuthenticationNotifier.new(
+        integration_id: integration_id,
```
jferris: Now that this is using composition, you can inline `"greenhouse"` here and remove `@integration_id` and its `attr_reader`, right?

app/jobs/sync_job.rb
base10: I figured if we were calling it multiple times, it made sense not to reinstantiate every time. If I'm only calling it once, no.

app/jobs/sync_job.rb
base10: Addressed in ef984da

app/jobs/sync_job.rb
base10: Addressed in ef984da

app/services/greenhouse/candidates_importer.rb
base10: Fixed with 6dfb25d

app/models/authentication_notifier.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for " \
+      "user_id: #{user.id} with #{integration_name}"
+    )
+  end
+
+  def default_exception
```
jferris: I think I would just eat the duplication here and pass `Unauthorized.new("Invalid authentication")` in multiple places.

base10: Fixed with 3c65278

app/models/authentication_notifier.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for " \
+      "user_id: #{user.id} with #{integration_name}"
+    )
+  end
+
+  def log_and_notify_of_unauthorized_exception(exception)
```
jferris: I think you can now reduce the public API of this class to only this method, right?

base10: Done in cb1653f

jferris:
Looks good!

base10:
Thank you. I appreciate the feedback to make this nicer.

PR #136, opened 2015-07-06
Refactor authorization notifications across integrations

base10:
* The ConnectionMailer now accepts an exception with changes upstream
  to match in User
* IntegrationAuthenticationNotification#set_exception now accepts an
  exception message
* Removed the unauthorized_notification from IcimsCandidateImportMailer
* Icims::CandidateImporter now uses ConnectionMailer to alert to
* authentication/authorization problems
* Icims::CandidateImporter is also using the
  IntegrationAuthenticationNotification module
  * More refactoring to happen here
* For: https://trello.com/c/NGozb2N4/

### DIFF
--- a/app/mailers/connection_mailer.rb
+++ b/app/mailers/connection_mailer.rb
@@ -1,6 +1,7 @@
 class ConnectionMailer < ApplicationMailer
-  def authentication_notification(connection_type:, email:)
+  def authentication_notification(connection_type:, email:, message:)
     @integration = map_connection_type_to_integration(connection_type)
+    @message = message
 
     mail(
       to: email,
--- a/app/models/icims/authorized_request.rb
+++ b/app/models/icims/authorized_request.rb
@@ -67,8 +67,12 @@ module Icims
       Rails.logger.info("ICIMS request END")
       r.execute
     rescue RestClient::Unauthorized => exception
-      user.send_connection_notification("icims")
-      raise Unauthorized, exception.message
+      unauthorized_exception = Unauthorized.new(exception.message)
+      user.send_connection_notification(
+        connection_type: "icims",
+        message: unauthorized_exception.message
+      )
+      raise unauthorized_exception
     end
 
     private
@@ -112,7 +116,5 @@ module Icims
     def hashed_payload
       digest.hexdigest(payload)
     end
-
-    class Unauthorized < StandardError; end
   end
 end
--- a/app/controllers/jobvite_imports_controller.rb
+++ b/app/controllers/jobvite_imports_controller.rb
@@ -1,7 +1,7 @@
 class JobviteImportsController < ApplicationController
   def create
     @jobvite_imports_presenter = ImportsPresenter.new(importer.import)
-  rescue Jobvite::Client::Unauthorized
+  rescue Unauthorized
     flash[:error] = t("jobvite_connections.authentication_error")
     redirect_to dashboard_path
   end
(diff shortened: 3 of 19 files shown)


base10:
Dependent on #135

app/mailers/connection_mailer.rb
```diff
@@ -1,5 +1,6 @@
 class ConnectionMailer < ApplicationMailer
-  def authentication_notification(connection_type:, email:)
+  def authentication_notification(connection_type:, email:, exception:)
```
jferris: What do you think about accepting a message here instead of an exception? The end user won't want other details from the exception, such as the class or backtrace.

base10: I'm good with that.

app/models/integration_authentication_notifications.rb
```diff
@@ -0,0 +1,27 @@
+module IntegrationAuthenticationNotifications
+  private
+
+  attr_reader :integration_id
+
+  def deliver_unauthorized_notification(exception = set_exception)
```
jferris: The number of default arguments in this module strike me as a smell. Is there meaningful duplication if you remove them? Can this method always accept an exception?

base10: In this PR, it can accept an exception and, I suppose, if I didn't have one to pass one it, I could establish one outside of this method and pass it in.

app/models/integration_authentication_notifications.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def log_unauthorized_exception(exception)
+    Rails.logger.error(
+      "#{exception.class} error #{exception.message} for " \
+      "user_id: #{user.id} with #{integration_name}"
+    )
+  end
+
+  def set_exception(error_message = "Invalid authentication")
```
base10: Here, I expect I could accept an error message and establish a default message elsewhere. My plan for accepting an error message externally was passing along meaningful exception information that might be coming in from the APIs themselves.

base10:
Future @jferris: Initial feedback addressed and ready for rereview.

app/models/icims/authorized_request.rb
```diff
@@ -67,8 +67,12 @@ def execute
       Rails.logger.info("ICIMS request END")
       r.execute
     rescue RestClient::Unauthorized => exception
-      user.send_connection_notification("icims")
-      raise Unauthorized, exception.message
+      unauthorized_exception = ::Unauthorized.new(exception.message)
```
jferris: Is the `::` prefix necessary here? If `Unauthorized` is ambiguous in some of our contexts, does it make sense for us to rename it?

base10: It's ambiguous currently, but I think we can remove the Icims::AuthorizedRequest::Unauthorized exception class.

base10: Fixed with 7250e48

app/services/greenhouse/candidates_importer.rb
```diff
@@ -15,6 +15,7 @@ def initialize(mailer, connection_repo, params, secret_key, signature)
       @secret_key = secret_key
       @connection_repo = connection_repo
       @candidate_name = CandidateName.new(greenhouse_payload)
+      @integration_id = "greenhouse"
```
jferris: This doesn't appear to be used.

base10: Fixed with f5c1f6b

jferris:
This looks good to me.

base10:
Thank you, @jferris. I'll let you know when I've got this merged in and have touched up #137 accordingly.

PR #137, opened 2015-07-07
Centralized candidate import mailing to a single object

base10:
* Adds CandidateImportMailer
* Removes IcimsCandidateImportMailer and
  GreenhouseCandidateImportMailer and their respective templates
* Adjusts the candidate import controllers and candidate import models
* Provides a retry link for unsuccessful imports of candidates
  from iCIMS
* Greenhouse::CandidateName now has a "#name" method, which acts the
  same as "#to_s" does on that method, to allow for the refactoring
* For: https://trello.com/c/NGozb2N4/ and https://trello.com/c/Wr0dqCQe/
* Depends on #135 and #136 prior to merge


app/helpers/candidate_import_mailer_helper.rb
```diff
@@ -0,0 +1,10 @@
+module CandidateImportMailerHelper
+  def retry_link
+    if @integration == "iCIMS"
```
jferris: This looks like integration-specific code in a generic helper. Some ideas to generalize it:

* Create a per-integration failure partial and render any integration-specific view code there.
* Create a `retriable?` method on the `Connection` subclasses so the generic code can decide whether or not the user is allowed to retry.

base10: Oh, smart. I like that.

base10: @jferris Fixed with 2a9e941. Coincidentally, a rereview would be appreciated.

app/mailers/candidate_import_mailer.rb
```diff
@@ -0,0 +1,31 @@
+class CandidateImportMailer < ApplicationMailer
+  def successful_import(candidate:, connection_type:, email:)
```
base10: Tempted to change `connection_type` to `integration_id` for consistency

app/views/candidate_import_mailer/successful_import.en.text.erb
```diff
@@ -0,0 +1,9 @@
+<%=
+  t(
+    "candidate_import_mailer.successful_import.message",
```
base10: Should be able to shorten this up to just `.message`

app/mailers/candidate_import_mailer.rb
jferris: Yeah, I think that would be an improvement.

app/services/greenhouse/candidates_importer.rb
```diff
@@ -15,6 +15,7 @@ def initialize(mailer, connection_repo, params, secret_key, signature)
       @secret_key = secret_key
       @connection_repo = connection_repo
       @candidate_name = CandidateName.new(greenhouse_payload)
+      @integration_id = "greenhouse"
```
jferris: Can this be inlined?

app/services/icims/candidate_importer.rb
```diff
@@ -6,6 +6,7 @@ def initialize(connection, mailer, params)
       @connection = connection
       @mailer = mailer
       @params = params
+      @integration_id = "icims"
```
jferris: Can we inline this, too?

jferris:
I had two minor comments. Looks good to merge after that.

app/services/greenhouse/candidates_importer.rb
base10: I have multiple places in the class where I'm using this (see 13ae792), so my pref would be to keep an accessor for this.

app/services/icims/candidate_importer.rb
base10: I have multiple places in the class where I'm using this (see 13ae792), so my pref would be to keep an accessor for this.

app/services/greenhouse/candidates_importer.rb
jferris: If you're worried about duplicating the value, what do you think about using a constant?

base10: I actually introduce one in #138 :D. I can do that here, though.

PR #138, opened 2015-07-07
Refactor *::CandidateImporter classes

base10:
Refactor *::Candidate(s)Importer classes

* CandidateImporter replaces two implementation specific classes
  * Greenhouse::CandidatesImporter
  * Icims::CandidateImporter
* CandidateImporter is a general interface to import candidates
* It relies on an import assistant to handle specifics and encapsulate what were
  the prior ::Candidate(s)Importer objects
* New import assistants will need to implement the following:
  * INTEGRATION_ID
    * e.g. "greenhouse" or "icims"
  * attribute_mapper
    * This could likely be delegated
  * candidate
  * import_candidate
  * initialize
  * skip_notification?
  * success?
* Whatever implementation specific needs exist can be addressed through private
  methods called within the public interface
* Should additional arguments be required for a new implementation, they can be
  passed in with "assistant_arguments"
  * See Greenhouse::CandidateImportAssistant for an example
* For: https://trello.com/c/Wr0dqCQe/

Follow-up items:

* Handle import controller `find_by!` calls differently when finding the
  connection
  * https://trello.com/c/ZFCTRBl3/
* Extract Greenhouse::CandidateImportAssistant's looking at ping requests and
  address that need elsewhere
  * If need be, pass that around as a separate object perhaps


app/services/common_candidate_importer.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(connection:, mailer:, params:)
+    @connection = connection
+    @integration_id = self.class::INTEGRATION_ID
+    @mailer = mailer
+    @params = params
+  end
+
+  private
+
+  def import_candidate
```
base10: Addressed with 1c4c45e

base10:
@jferris, the accumulated feedback from #135, #136, and #137 has been applied here. Ready for review.

app/controllers/greenhouse_candidate_imports_controller.rb
```diff
@@ -4,11 +4,11 @@ class GreenhouseCandidateImportsController < ApplicationController
 
   def create
     Greenhouse::CandidatesImporter.new(
-      CandidateImportMailer,
-      Greenhouse::Connection,
-      greenhouse_candidate_import_params,
-      params['secret_key'],
-      request.headers['Signature']).import
+      mailer: CandidateImportMailer,
```
base10: :abc:

app/services/common_candidate_importer.rb
```diff
@@ -0,0 +1,54 @@
+module CommonCandidateImporter
```
base10: Thoughts on this being a module and mixed in vs. making it a composable instead?

app/controllers/greenhouse_candidate_imports_controller.rb
```diff
@@ -17,6 +17,12 @@ def create
 
   private
 
+  def connection
+    Greenhouse::Connection.find_by(
```
jferris: Would `find_by!` be better here? It doesn't look like we handle the case where `find_by` returns `nil`.

app/services/common_candidate_importer.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def namely_importer
+    NamelyImporter.new(
+      attribute_mapper: attribute_mapper,
+      namely_connection: user.namely_connection,
+    )
+  end
+
+  def notifier
```
base10: Potentially better as `#authentication_notifier` now

app/controllers/greenhouse_candidate_imports_controller.rb
base10: Oh, good call.

app/services/icims/candidate_importer.rb
```diff
@@ -1,44 +1,27 @@
...
-        mailer.delay.unsuccessful_import(
-          candidate: candidate,
-          email: user.email,
-          integration_id: INTEGRATION_ID,
-          status: imported_result
-        )
-      end
+      import_candidate
     rescue Icims::Client::Error => exception
       notifier.log_and_notify_of_unauthorized_exception(exception)
```
base10: If this moves to a class, this would be an excellent thing to delegate in said class.

app/services/common_candidate_importer.rb
jferris: I think it would likely work better as a class. It may actually make sense to flip the composition, so that we have a generic `Import` type which composes a strategy for Greenhouse/ICIMS.

app/services/icims/candidate_importer.rb
```diff
@@ -1,44 +1,27 @@
...
     end
 
+    def imported_result
+      @imported_result ||= namely_importer.single_import(candidate)
```
base10: Check and see if this actually needs to be memoized.

app/services/common_candidate_importer.rb
base10: Interesting idea. Will ponder that for a few minutes.

app/controllers/greenhouse_candidate_imports_controller.rb
base10: Fixed with: 180474f

app/controllers/greenhouse_candidate_imports_controller.rb
base10: Fixed with 180474f

app/services/icims/candidate_importer.rb
mike-burns: After the module -> class refactoring, you might be able to drop the memoization more easily.

app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -3,16 +3,18 @@ class IcimsCandidateImportsController < ApplicationController
...
+    ).import
     render nothing: true
   end
 
   private
 
   def connection
-    Icims::Connection.find_by(api_key: params[:api_key],
+    Icims::Connection.find_by!(api_key: params[:api_key],
                               customer_id: params[:customerId])
```
mike-burns: This indentation is off.

app/controllers/greenhouse_candidate_imports_controller.rb
```diff
@@ -17,6 +17,12 @@ def create
 
   private
 
+  def connection
+    Greenhouse::Connection.find_by!(
```
mike-burns: Using `#find_by!` here leaks information; it should raise `Unauthorized` instead. Using `#find_by!` here means the exploiter will see confirmation on whether they used the correct secret key.

app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -3,16 +3,18 @@ class IcimsCandidateImportsController < ApplicationController
...
+      params: params
+    ).import
     render nothing: true
   end
 
   private
 
   def connection
-    Icims::Connection.find_by(api_key: params[:api_key],
+    Icims::Connection.find_by!(api_key: params[:api_key],
```
mike-burns: Does this show a useful error message to the user? Conversely, does it leak information to an attacker? Is there a more accurate result we can show than the default RecordNotFound page?

(My bias: I don't love `#find_by!`.)

app/controllers/greenhouse_candidate_imports_controller.rb
base10: Good to know, I was basing the change off of similar feedback I got on a separate PR.

app/services/icims/candidate_import_assistant.rb
```diff
@@ -0,0 +1,48 @@
+module Icims
+  class CandidateImportAssistant
+    INTEGRATION_ID = "icims"
+
+    delegate :connection, to: :context
+    delegate :params, to: :context
+
+    def initialize(assistant_arguments: {}, context:)
```
base10: I don't feel great about it, but I'd prefer to have the interfaces between `Icims::CandidateImportAssistant` and `Greenhouse::CandidateImportAssistant` the same.

mike-burns: Can you name it with an underscore to further indicate that it is unused? This might be imposible with kw args.

base10: There's `**` and `**_` apparently: http://flushentitypacket.github.io/ruby/2015/03/31/ruby-keyword-arguments-the-double-splat-and-starsnake.html

base10: Fixed in 8c5d094. Started with `**`, but looks like `**_` is a better fit since it'd otherwise be something like `**other_arguments`

app/controllers/icims_candidate_imports_controller.rb
base10: Since this PR has already taken a long and winding road, I'd like to tackle this feedback as a separate card: https://trello.com/c/ZFCTRBl3/

app/controllers/greenhouse_candidate_imports_controller.rb
base10: I'd like to address this in a separate PR. Tracking card here: https://trello.com/c/ZFCTRBl3/

mike-burns:
This is a long-running branch! I hope you won't hate me if I leave some comments. Do with them as you please:

- Instead of `Greenhouse::CandidateImportAssistant`, it could be named `Greenhouse::CandidateImporter`. It's shorter and describes what it does more succinctly.
- Instead of `assistant_class`, it could be named `importer_factory`. That's what it is: it's a factory method/object. But,
- A cool thing about factory objects is that they have a tiny interface. Your `assistant_class` objects need to respond to `::INTEGRATION_ID`. How about moving that into the instantiated object instead: `import_assistant.integration_id` (or, `importer.integration_id`)? Then the factory can be anything that responds to `.new`, including a simple `double`(this might simplify your testing life).

base10:
@mike-burns Thanks for the feedback! I have some context that I'd appreciate some clarification or further color on:

* With `::CandidateImportAssistant` I want to make clear that this class is assisting something else vs. being a standalone. I also wanted to keep it separate from the in place `::CandidateImporter` while I worked this out.
* `assistant_class` started out as `strategy_class` and I hadn't really been thinking of it as a factory, per se.
* I'm fine with moving `::INTEGRATION_ID` in, but @jferris provided me feedback to _not_ do that.

I realize there's some manner of opinion differences here, but does that shade anything differently for you?

base10:
@mike-burns Also, yeah, long running PR because I did four in a row while Joe was out of office and addressed feedback in them iteratively. Getting from opening the PR to here has been an interesting path. :D

mike-burns:
Works for me! LGTM.

base10:
Awesome, thank you.

PR #139, opened 2015-07-08
Split Authorizations from Connections

jferris:
This separates the concept of authenticating with an integration from
the concept of being connected. This us allows us to separately
create/update credentials. It also clarifies the reason for using form
objects for connection initiation.


base10:
@jferris Overall, I'm fine with the code, but I'm not sure I'm mentally following the desire to separate the concepts of connection and authentication. Anything that might help me picture it the way you are?

jferris:
@base10 the original motivation is for https://trello.com/c/xrYEiY3u.

We need to handle three forms:

* Authorize a new integration (before: `connections#new`, after: `authorizations#new`)
* Fill in additional fields after authorization (before and after: `connections#edit`)
* Edit authorization

For the last form, it's not clear where this would go, as there's only one controller, and "edit" is already taken. Stepping back from this, I felt like we were conflating the concerns of authorization and configuration, which is why I couldn't find a place to add the new form.

After this pull request, I can implement the last form as `authorizations#edit`.

base10:
@jferris: Good with change. Separately, are we good with calling this "authorization" vs. "authentication"?

jferris:
> Good with change. Separately, are we good with calling this "authorization" vs. "authentication"?

Good question. The forms collect login details, so I guess "authentication" is likely a better term.

base10:
LGTM

PR #140, opened 2015-07-09
Users can edit connection credentials

jferris:
* Adds edit links for each connection
* Implements edit/update for authentications

https://trello.com/c/xrYEiY3u

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3,6 +3,7 @@ en:
     show:
       connect: "Connect"
       disconnect: "Disconnect"
+      edit: "Edit"
       export_now: "Export now"
       import_now: "Import now"
       sign_out: "Sign out"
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -154,6 +155,10 @@ en:
         title: "Greenhouse"
         slogan: >
           Please fill out the following fields to connect to this application.
+      edit:
+        title: "Greenhouse"
```
base10: Thoughts on refactoring the duplication out between "new" and "edit" here? Potentially using `integration_id.name" instead?

base10:
One question, otherwise LGTM

config/locales/en.yml
jferris: I've been thinking we probably don't need any variation between these pages, and the extra translations are probably just noise.

What do you think about doing a separate pull request to:

* Use the integration title for all these titles
* Use the same slogan regardless of integration

We're already doing both of those, but we repeat the translations anyway.

base10: I like that idea.

PR #141, opened 2015-07-09
Consolidate translations for page titles

jferris:
We were repeating the same headings for each integration. This
consolidates the headings into a single key for each page, reducing
noise in pull requests and reducing the number of steps required to add
a new integration.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -148,64 +148,38 @@ en:
       connected_html: Connected NetSuite Description
       disconnected_html: Disconnected NetSuite Description
 
+  authentications:
+    new:
+      title: "Connect to %{integration}"
+      slogan: >
+        Please fill out the following fields to connect to this application.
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -148,64 +148,38 @@ en:
...
+        Please fill out the following fields to update your credentials.
+
+  connections:
+    edit:
+      title: "Connect to %{integration}"
+      slogan: >
+        Please fill out the following fields to finalize this connection.
+  syncs:
+    create:
+      title: "Exporting to NetSuite"
```
base10: Do we want `%{integration}` here?

base10:
One nit, otherwise LGTM.

config/locales/en.yml
jferris: Whoops, fixed in 1aaa617.

PR #142, opened 2015-07-09
Refactor spec to avoid instance state

jferris:
Removes hidden dependencies and temporal coupling.

https://trello.com/c/58rvUSyp


base10:
LGTM

PR #143, opened 2015-07-14
Jobvite is a sync

mike-burns:
- JobviteConnection responds to #sync, which means it can use the
  asynchronous SyncController framework.
- Remove unused JobviteImportsController and associated trimmings.
- Remove unused app/views/net_suite_exports/_export.html.erb template.
- ImportResult is an Enumerable, which means we can skip calling #to_a.
- Move knowledge of how to combine first and last names into the models.

https://trello.com/c/fS4LrFwi


app/models/jobvite/connection.rb
```diff
@@ -30,5 +30,42 @@ def disconnect
...
+
+    def sync
+      import_results.map { |result| Result.new(**result) }
+    end
+
+    private
+
+    def import_results
+      Importer.new(
+        user,
```
base10: Thoughts on passing user in with a named argument?

spec/models/import_result_spec.rb
```diff
@@ -33,19 +33,23 @@
...
       result[candidate_one] = :failure
       result[candidate_two] = :success
 
-      expect(result.to_a).to eql [{:candidate=>{:first_name=>"Mike"},
-                                   :result=>:failure},
-                                   {:candidate=>{:first_name=>"Nolan"},
-                                    :result=>:success}]
+      actual = []
+      result = described_class.new(attribute_mapper_double)
+      result.each { |c_r| actual << c_r }
```
base10: Thoughts on expanding `c_r` to `candidate` or similar?

app/models/import_result.rb
```diff
@@ -15,9 +17,9 @@ def to_s(format = "%{candidate}: %{result}\n")
     end.join
   end
 
-  def to_a
-    results.map do |candidate, result|
-      { candidate: candidate, result: result }
+  def each(&block)
```
base10: This looks like a good learning opportunity for me. What's the benefit of defining `each` here vs. `to_a`? Getting this to behave more like an Enumerable?

base10:
The removal of `app/views/net_suite_exports/_export.html.erb` looks fine since NetSuite exports now run through the SyncsController.

base10:
A few questions/suggestions, but I like where this is leading.

app/models/import_result.rb
mike-burns: Yes; the contract for an Enumerable is that the object must define `#each`. Additionally, if you want to use `#sort`, `#max`, or `#min`, the object must define `#<=>`.

By making this Enumerable we can skip calling `#to_a` explicitly before calling e.g. `#select` or `#each`.

app/models/jobvite/connection.rb
mike-burns: Good thinking.

spec/models/import_result_spec.rb
mike-burns: Thanks for the name suggestion; it's much better than `c_r`.

mike-burns:
Is this ready to merge?

base10:
Let me look through the latest changes quick.

base10:
Left a couple of :abc: comments on a506bd6, otherwise LGTM.

base10:
@mike-burns Oh, hey the build's failing, that'd be good to address :D

PR #144, opened 2015-07-15
Missing secrets are blank 401

mike-burns:
Currently, a missing API or secret key shows a different error message than other authentication failures. This leaks data to an attacker: if someone is trying to guess secret keys, they will know when they've succeeded because the error message will change.

Instead, show the same error message for both a missing key and an incorrect key.

This branch took some turns; feel free to look through the commits to see other ideas. Some might be better than others.

While here, clean up some of the syntax in some related tests.


app/controllers/icims_candidate_imports_controller.rb
```diff
@@ -10,14 +10,16 @@ def create
...
 
   private
 
   def connection
-    Icims::Connection.find_by!(
+    Icims::Connection.find_by(
       api_key: params[:api_key],
       customer_id: params[:customerId]
-    )
+    ) or raise Unauthorized.new(Unauthorized::DEFAULT_MESSAGE)
```
mike-burns: Hound, I love ya, but we both know that using `||` is invalid syntax.

spec/requests/greenhouse_statuses_spec.rb
```diff
@@ -7,19 +7,19 @@
...
   end
-  let(:connection) do 
-    create(:greenhouse_connection, :with_namely_field, name: "myhook") 
+  let(:connection) do
+    create(:greenhouse_connection, :with_namely_field, name: "myhook")
   end
 
   before do
-    stub_request(:get, /#{ api_host }\/api\/v1\/profiles\/fields/).
+    stub_request(:get, %r{#{ api_host }/api/v1/profiles/fields}).
```
base10: :heart_eyes:

spec/requests/greenhouse_statuses_spec.rb
```diff
@@ -77,17 +84,17 @@
     )
   end
 
-  def greenhouse_ping
-    @greenhouse_ping ||= JSON.parse(
+  let(:greenhouse_ping) do
```
base10: Since these are `let`s know, does it make any sense to move them above the tests? (I ask because that's at least where I'm accustomed to seeing most `let` and `subject` setting in spec tests.

mike-burns: Ah, I do not know the style of this project, so you'll have to tell me. Looking at other tests, it does seem that they are grouped at the top. I can move these.

I always put them at the bottom with the private methods because that's all they are (as demonstrated in this PR).

base10:
So, one thing I'd imagined when writing the card for this was instead of getting an exception, we'd bounce the user back to the dashboard to adjust their connection. Is it better in cases like this to not be helpful like that? I'd expect we'd still keep the messaging neutral about missing vs. incorrect.

mike-burns:
I wonder about the candidates importers overall: should they be asynchronous? Should they use `SyncJob`?

base10:
I think moving them to be asynchronous would be a good move. I think whether or not `SyncJob` is the right fit would be a follow-up question.

mike-burns:
Good discussion; I've updated the Trello card (which I should have linked to in the PR description; apologies).

Closing this PR so that I can open a substantially different one later.

PR #145, opened 2015-07-21
First pass at AttributeMapper and FieldMapping objects

base10:
* AttributeMapper belongs to a User, primarily so if a User goes away, the
  AttributeMapper will, too
* AttributeMappers will be created per connection. AttributeMappers have_a
  connection, Connections belong_to an AttributeMapper
  * This is in order to avoid needing a polymorphic relationship on
    AttributeMapper
* NB: Not yet wired into NetSuite::Export
* NB: Does not yet handle fields that need any sort of additional handling

Following a recommendation from Mike Burns to not create this implicitly by
calling `#attribute_mapper`, particularly when a GET request uses
NetSuite::Connection, it may have database side-effects, which violates the
Principle of Least Surprise.

We're not thrilled with having it in an `after_create` here, but we like it
better than it being implicit.

* For: https://trello.com/c/JsdID6YF/

### DIFF
--- a/db/migrate/20150715184131_create_attribute_mappers.rb
+++ b/db/migrate/20150715184131_create_attribute_mappers.rb
@@ -1,0 +1,10 @@
+class CreateAttributeMappers < ActiveRecord::Migration
+  def change
+    create_table :attribute_mappers do |t|
+      t.integer :mapping_direction, default: 0, null: false
+      t.references :user, foreign_key: true, index: true, null: false
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/app/models/attribute_mapper.rb
+++ b/app/models/attribute_mapper.rb
@@ -1,0 +1,19 @@
+class AttributeMapper < ActiveRecord::Base
+  belongs_to :user, dependent: :destroy
+  has_many :field_mappings
+
+  validates :mapping_direction, presence: true
+  validates :user, presence: true
+  validates :user_id, presence: true
+
+  enum mapping_direction: { import: 0, export: 1 }
+
+  def build_field_mappings(default_field_mapping)
+    default_field_mapping.each_pair do |namely_field, integration_field|
+      field_mappings << FieldMapping.new(
+        integration_field_name: integration_field.to_s,
+        namely_field_name: namely_field
+      )
+    end
+  end
+end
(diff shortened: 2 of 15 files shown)


base10:
@mike-burns provided an initial round of feedback from Slack https://thoughtbot.slack.com/archives/namely/p1437478523000046:

> Cool. Some initial questions, though it is still a WIP: Does NetSuite::AttributeMapperBuilder need to expose #attribute_mapper? Can NetSuite::AttributeMapperBuilder#build_field_mappings be moved into AttributeMapper? Should NetSuite::Connection be the object that creates its attribute mapper? Should it be so implict?

> This is a direction that I like.

> Oh, does it make sense to use an enum for mapping_direction?

> Don't be afraid to use #let.

db/migrate/20150715184131_create_attribute_mappers.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAttributeMappers < ActiveRecord::Migration
+  def change
+    create_table :attribute_mappers do |t|
+      t.integer :mapping_direction, default: 0, null: false
```
mike-burns: Default 0?

base10: I figured from reading the enum docs that defaulting would be a good idea. This will be treated as an `:import`. I take it you'd prefer to not set a default and if it's not set, let it fail?

mike-burns:
Cool. I'm interested to see how this will connect with the rest of the system -- currently nothing uses this code?

base10:
That's correct. We need to have an initial idea about how we want to handle fields like subsidiary that come from a configuration and gender that needs to be mapped and then we can get this wired into `NetSuite::Export` to complete the first step.

db/migrate/20150715184131_create_attribute_mappers.rb
mike-burns: Mostly curious why there is a default. I'm not sure if any default makes sense here, but I'm also not opposed. Thanks for the explanation.

base10:
@mike-burns Good to merge and integrate, then?

mike-burns:
Yup.

PR #146, opened 2015-07-22
Perform a very rudimentary profile conversion

base10:
* Like Jobvite::AttributeMapper, non-existent values won't be converted
* This is not yet doing anything about fields that need special handling
* I expect that I'll need to do further work with
  NetSuite::AttributeMapper for that special handling, which will
  likely include passing NetSuite::AttributeMapper to
  ::AttributeMapper as context for special handling callbacks
* I'm ambivalent about `call` vs. `import` and `export`, but I
  wanted to see how it felt to use `mapping_direction` with this from
  ::AttributeMapper
* I suspect there is some opportunity for more succinct expressions
  here, feedback welcome
* For: https://trello.com/c/JsdID6YF/

### DIFF
--- a/app/models/attribute_mapper.rb
+++ b/app/models/attribute_mapper.rb
@@ -16,4 +16,22 @@ class AttributeMapper < ActiveRecord::Base
       )
     end
   end
+
+  def call(profile)
+    send(mapping_direction, profile)
+  end
+
+  private
+
+  def export(profile)
+    field_mappings.each_with_object({}) do |field_mapping, accumulator|
+      if profile[field_mapping.namely_field_name].present?
+        accumulator.merge!(
+          field_mapping.integration_field_name => profile[
+            field_mapping.namely_field_name
+          ]
+        )
+      end
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


base10:
Next up for this:

* Handling fields that need special handling
* Bringing this into `NetSuite::Export`

app/models/attribute_mapper.rb
```diff
@@ -16,4 +16,22 @@ def build_field_mappings(default_field_mapping)
...
+    converted_profile = {}
+    field_mappings.each do |field_mapping|
+      if profile[field_mapping.namely_field_name].present?
+        converted_profile[field_mapping.integration_field_name] =
+          profile[field_mapping.namely_field_name]
+      end
+    end
+
+    converted_profile
+  end
```
mike-burns: You can shorten this by two lines with `Enumerable#each_with_object`:

```ruby
def export(profile)
  field_mappings.each_with_object({}) do |field_mapping, acc|
    if profile[field_mapping.namely_field_name].present?
      acc.merge!(
        field_mapping.integration_field_name => profile[field_mapping.namely_field_name]
      )
    end
  end
end
```

mike-burns:
My hunch is that we will replace `#call` with `#export` entirely and drop the `mapping_direction`, but I'm interested to see what `NetSuite::Export` will look like with this before making any, er, demands.

I have a thought on using `#each_with_object` instead of `#each`, and otherwise: LGTM.

app/models/attribute_mapper.rb
base10: Oh, an accumulator. Neat. I just learned that those exist.

PR #147, opened 2015-07-23
Edit the field mapping

mike-burns:
This is the front-end code for editing the field mappings for Netsuite. It does not work all the way through: the Netsuite export does not make use of the field mappings yet.

This PR serves as a preview of what's to come, and as a first pass.

https://trello.com/c/rgC5mN3g/97-ui-for-editing-a-netsuite-field-mapping


app/controllers/mappings_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    else
+      render "edit"
+    end
+  end
+
+  private
+
+  def attribute_mapper_params
+    params.require(:attribute_mapper).
+      permit(field_mappings_attributes: [:namely_field_name, :id])
```
base10: We'd need `:integration_field_name` here as well, yes?

mike-burns: That's not a field in the form. Rather, the `namely_field_name` is tracked by its index in the array of `field_mappings` on an `AttributeMapper` (using `accepts_nested_attributes_for`).

jferris:
I'm closing this while Nathan finishes up #148 and I connect the attribute mapper from his branch with the form and export class from this branch.

PR #148, opened 2015-07-27
NetSuite::AttributeMapper can provide export-ready attributes

base10:
* NetSuite::AttributeMapper will be passed as context to
  ::AttributeMapper
* NetSuite::AttributeMapper defines a hash in `#post_mapping_handlers`
  * These will convert a matching value in the transformed profile to
    what will be expected at NetSuite
  * I didn't implement this as an `inject` since I'm not always
    expecting to have a value in the transformed profile, but I want to
    ensure one exists after post handling
* This will need to be hooked into
  `NetSuite::Export::Employee#attributes`
* This will replace handling in NetSuite::Export for mapping gender,
  handling the job title data structure and setting the subsidiary id
* For: https://trello.com/c/at7jXc36/

### DIFF
--- a/app/models/attribute_mapper.rb
+++ b/app/models/attribute_mapper.rb
@@ -2,11 +2,8 @@ class AttributeMapper < ActiveRecord::Base
   belongs_to :user, dependent: :destroy
   has_many :field_mappings
 
-  validates :mapping_direction, presence: true
   validates :user, presence: true
   validates :user_id, presence: true
-
-  enum mapping_direction: { import: 0, export: 1 }
 
   def build_field_mappings(default_field_mapping)
     default_field_mapping.each_pair do |namely_field, integration_field|
@@ -17,19 +14,13 @@ class AttributeMapper < ActiveRecord::Base
     end
   end
 
-  def call(profile)
-    send(mapping_direction, profile)
-  end
-
-  private
-
   def export(profile)
     field_mappings.each_with_object({}) do |field_mapping, accumulator|
-      if profile[field_mapping.namely_field_name].present?
+      if profile.send(field_mapping.namely_field_name).present?
         accumulator.merge!(
-          field_mapping.integration_field_name => profile[
+          field_mapping.integration_field_name => profile.send(
             field_mapping.namely_field_name
-          ]
+          )
         )
       end
     end
--- a/app/models/net_suite/attribute_mapper.rb
+++ b/app/models/net_suite/attribute_mapper.rb
@@ -1,14 +1,46 @@
 class NetSuite::AttributeMapper
-  delegate :call, to: :attribute_mapper
+  GENDER_MAP = {
+    "Male" => "_male",
+    "Female" => "_female",
+    "Not specified" => "_omitted",
+  }
+  GENDER_MAP.default = "_omitted"
+
   delegate :field_mappings, to: :attribute_mapper
   delegate :mapping_direction, to: :attribute_mapper
+  delegate :persisted?, to: :attribute_mapper
 
   def initialize(attribute_mapper:, configuration:)
     @attribute_mapper = attribute_mapper
     @configuration = configuration
   end
 
+  def export(profile)
+    attribute_mapper.export(profile)
+  end
+
+  def post_handle(exported_profile)
+    exported_profile["gender"] = map_gender(exported_profile["gender"])
+    exported_profile["subsidiary"] = set_subsidiary_id
+    exported_profile["title"] = format_job_title(exported_profile["title"])
+
+    exported_profile
+  end
+
   private
+
+  def format_job_title(value)
+    value = Hash(value)
+    value.fetch(:title) { "" }
+  end
+
+  def map_gender(value)
+    GENDER_MAP[value]
+  end
+
+  def set_subsidiary_id
+    { "internalId" => configuration.subsidiary_id }
+  end
 
   attr_reader :attribute_mapper, :configuration
 end
--- a/spec/models/net_suite/attribute_mapper_spec.rb
+++ b/spec/models/net_suite/attribute_mapper_spec.rb
@@ -1,12 +1,13 @@
 require "rails_helper"
 
 describe NetSuite::AttributeMapper do
+  let(:configuration) { double("configuration", subsidiary_id: "123") }
   let(:netsuite_attribute_mapper) do
     NetSuite::AttributeMapper.new(
       attribute_mapper: NetSuite::AttributeMapperBuilder.new(
         user: create(:user),
       ).build,
-      configuration: {}
+      configuration: configuration
     )
   end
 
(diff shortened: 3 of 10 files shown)


app/models/attribute_mapper.rb
```diff
@@ -17,21 +17,36 @@ def build_field_mappings(default_field_mapping)
     end
   end
 
-  def call(profile)
-    send(mapping_direction, profile)
+  def call(context:, profile:)
+    @context = context
+    @profile = profile
+    @mapped_profile = send(mapping_direction)
```
jferris: Instead of using `send` here, what about having both directions be public methods and calling them directly?

spec/models/net_suite/attribute_mapper_spec.rb
```diff
@@ -21,45 +22,160 @@
...
-      "last_name" => "Last"
+      job_title: {
+        id: "1234",
+        title: title
+      }
     }
   end
+
+  def stubbed_profile(data = stubbed_profile_data)
+    OpenStruct.new(data)
```
jferris: Thoughts on using `double` instead of `OpenStruct`? Using `OpenStruct` will also define all attributes and writers, regardless of whether they will exist in production.

spec/models/net_suite/attribute_mapper_spec.rb
```diff
@@ -21,45 +22,160 @@
...
+      )
 
       expect(deleted_import_fields.to_set).not_to be_subset(
-        export_ready_profile.keys.to_set
+        export_attributes.keys.to_set
       )
     end
+
+    describe "value handling" do
+      it "sets expected values in the profile for regular attributes" do
```
jferris: Do you think it's worth writing out the duplicated values in the setup and unique values in the expectation to be explicit and simple in the test? It's definitely a fuzzy line, but using computations like `to_set` make it harder to know what the actual test condition is and whether or not it's being properly executed.

app/models/attribute_mapper.rb
base10: Mike Burns and I talked about that some. I'm not opposed. That'll mean dropping `#mapping_directon` since it's not likely useful, otherwise.

spec/models/net_suite/attribute_mapper_spec.rb
base10: Yeah, that seems fine.

spec/models/net_suite/attribute_mapper_spec.rb
base10: I see what you're getting at. Let me see how that cleans things up.

base10: Addressed in a889e77

spec/models/net_suite/attribute_mapper_spec.rb
base10: Fixed with 96fd077

app/models/attribute_mapper.rb
base10: Initial piece addressed in d459188. Dropping `mapping_direction` next.

base10: And `mapping_direction` is dropped in 753637a

base10:
@jferris Ready for rereview.

app/models/attribute_mapper.rb
```diff
@@ -17,21 +14,33 @@ def build_field_mappings(default_field_mapping)
...
+          field_mapping.integration_field_name => profile.send(
             field_mapping.namely_field_name
-          ]
+          )
         )
       end
     end
   end
+
+  def post_handling(context:, mapped_profile:)
```
jferris: What do you think about defining `post_handle` on each connection object that accepts an unmapped profile and returns a mapped profile? This would:

* Shrink the mapping API from `post_mapping_handlers` and then a method for each handler (N+1) to `post_handle` (1)
* Allow each integration strategy to do anything it needs for post-processing rather than being restricted to value mapping
* Hide the per-integration mapper object within the connection

base10: I generally follow, but I'm not understanding the benefit of having `post_handle` on the connection object instead of the attribute mapper. My thinking was that when I called `NetSuite::AttributeMapper#call`, I had exactly what I need to pass along from `NetSuite::Export::Employee#attributes` (or a further refactoring)

jferris: My thinking is that instead of enforcing a family of classes for each integration (Connection, AttributeMapper, Client), it would be wise to have a single, public-facing class (Connection) for each integration. We can compose smaller classes as necessary to keep the connection classes small, but it prevents all the little classes getting tied up in the more generic layers like controllers, the attribute mapper, and background jobs.

base10: I've been flipping this around in my head for a few minutes. I'll give it a shot. I'm expecting that `NetSuite::AttributeMapper` will go away entirely in that case.

jferris: It may still be valuable to provide state and keep `Connection` at a reasonable size, but it's hard to know until after the refactor. I'd refator to composition first and inline the class second if it feels like it's just adding complexity.

base10:
@jferris Revised. I'm now setting-up `NetSuite::Connection` to handle `#export` and `#post_handle`. These are delegated to `NetSuite::AttributeMapper`. I'm open to pulling this functionality into `NetSuite::Connection`, but I think I'd like to explore that in a separate PR.

My expectation is we'd adjust `NetSuite::Export` to call `#export` and `#post_handle` within `NetSuite::Export::Employee#attributes` as a first step, since that should be a clean replacement, also as a separate PR.

Thoughts?

app/models/net_suite/attribute_mapper.rb
```diff
@@ -1,14 +1,57 @@
...
     @attribute_mapper = attribute_mapper
     @configuration = configuration
   end
 
+  def export(profile)
+    attribute_mapper.export(profile)
+  end
+
+  def post_handle(exported_profile)
+    post_mapping_handlers.each_pair do |attribute, method|
```
jferris: Since `post_mapping_handlers` is a list of private methods on this class, do you think it would be simpler to call those methods directly instead?

base10: I sure can.

jferris:
> I'm open to pulling this functionality into NetSuite::Connection, but I think I'd like to explore that in a separate PR.

Yeah, I think it's fine to keep that in place for this change.

app/models/net_suite/attribute_mapper.rb
base10: Addressed with 3f994d8

app/models/net_suite/attribute_mapper.rb
```diff
@@ -1,14 +1,57 @@
...
+    end
+
+    exported_profile
+  end
+
   private
 
+  def format_job_title(value)
+    value = Hash(value)
+    value.fetch(:title) { "" }
```
jferris: This is making an assumption on the data source (Namely's `job_title` field) and not the target (NetSuite's title field). This won't cause issues yet, because you can't change mappings, but it will when we allow users to map other fields into the `title` field.

For example, if I map a custom text field from Namely into the `title` field in NetSuite, this will cause a crash, because it tries to treat the text value as a `job_title` object.

I think it probably makes sense to merge this anyway, but we should keep in mind what assumptions we're making about types and try to limit them where possible.

base10: Good point. Will need to think on that a bit.

jferris:
This looks good to merge.

PR #149, opened 2015-07-28
Users can map NetSuite fields to Namely fields

jferris:
Because:

* Not all fields are present in all integrations
* Not all integration fields are supported by Namely
* We need to know which integration fields map to which Namely fields

This commit:

* Introduces field mapping form when setting up a new connection
* Uses the new FieldMapper class to map field names and values

Notes:

* Currently, only NetSuite connections support field mapping

https://trello.com/c/M2HMOphf

### DIFF
--- a/app/controllers/connections_controller.rb
+++ b/app/controllers/connections_controller.rb
@@ -5,7 +5,7 @@ class ConnectionsController < IntegrationController
 
   def update
     if update_connection
-      redirect_to dashboard_path
+      redirect_to after_update_path
     else
       render :edit
     end
@@ -24,5 +24,13 @@ class ConnectionsController < IntegrationController
 
   def connection_params
     params.require(connection_type).permit(connection.allowed_parameters)
+  end
+
+  def after_update_path
+    if connection.attribute_mapper?
+      edit_integration_mapping_path(integration_id)
+    else
+      dashboard_path
+    end
   end
 end
--- a/app/models/attribute_mapper.rb
+++ b/app/models/attribute_mapper.rb
@@ -5,6 +5,52 @@ class AttributeMapper < ActiveRecord::Base
   validates :user, presence: true
   validates :user_id, presence: true
 
+  accepts_nested_attributes_for :field_mappings
+
+  NAMELY_FIELDS = [
+    :asset_management,
+    :bio,
+    :corporate_card_number,
+    :current_job_description,
+    :dental_info,
+    :departure_date,
+    :dob,
+    :email,
+    :emergency_contact,
+    :emergency_contact_phone,
+    :employee_handbook,
+    :employee_id,
+    :employee_wage_theft_prevention_act,
+    :first_name,
+    :gender,
+    :healthcare_info,
+    :home_phone,
+    :image,
+    :job_change_reason,
+    :job_description,
+    :job_title,
+    :key_tag_number,
+    :laptop_asset_number,
+    :last_name,
+    :life_insurance_info,
+    :linkedin_url,
+    :marital_status,
+    :middle_name,
+    :mobile_phone,
+    :office_company_mobile,
+    :office_direct_dial,
+    :office_fax,
+    :office_main_number,
+    :office_phone,
+    :personal_email,
+    :preferred_name,
+    :resume,
+    :start_date,
+    :test_custom_field,
+    :user_status,
+    :vision_plan_info,
+  ]
+
   def build_field_mappings(default_field_mapping)
     default_field_mapping.each_pair do |namely_field, integration_field|
       field_mappings << FieldMapping.new(
--- a/app/models/net_suite/attribute_mapper.rb
+++ b/app/models/net_suite/attribute_mapper.rb
@@ -16,22 +16,18 @@ class NetSuite::AttributeMapper
   end
 
   def export(profile)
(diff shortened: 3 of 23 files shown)


app/controllers/connections_controller.rb
```diff
@@ -25,4 +25,12 @@ def update_connection
   def connection_params
     params.require(connection_type).permit(connection.allowed_parameters)
   end
+
+  def after_update_path
```
base10: Is it worth extracting this and `AuthenticationsController#after_ready_path` to a common method or does that make things less obvious than they should be?

jferris: It's sort of a wizard, so they go to different places depending on the stage.

This method should actually go away once all the integrations support attribute mapping.

jferris:
This is a non-work-in-progress edition of #147.

app/models/net_suite/export.rb
```diff
@@ -73,15 +73,7 @@ def id
...
-        {
-          firstName: @profile.first_name,
-          lastName: @profile.last_name,
-          email: @profile.email,
-          gender: map_gender(@profile.gender),
-          phone: @profile.home_phone,
-          subsidiary: { internalId: @configuration.subsidiary_id },
-          title: @profile.job_title[:title]
-        }
+        @attribute_mapper.export(@profile)
```
base10: We'll also need to call `@attribute_mapper.post_handle` here.

base10:
In `NetSuite::Export::Employee` we can drop the gender mapping pieces and job title handling.

app/models/net_suite/export.rb
jferris: Hmm, it's surprising that none of the tests failed. I'll expand the coverage.

jferris:
> In NetSuite::Export::Employee we can drop the gender mapping pieces and job title handling.

I removed `gender_map` in eaf8227. I'm not seeing any code for job title handling. Can you paste a link?

app/models/attribute_mapper.rb
```diff
@@ -5,6 +5,52 @@ class AttributeMapper < ActiveRecord::Base
   validates :user, presence: true
   validates :user_id, presence: true
 
+  accepts_nested_attributes_for :field_mappings
+
+  NAMELY_FIELDS = [
```
base10: This will work for now, but we should follow-up and ask the Namely API for the fields it knows about. See https://developers.namely.com/docs/fields-index

jferris: That's in progress on another branch: https://github.com/namely/connect/commit/9965688b13c8b385c38aefb82d23d9e05a7e1c31

app/models/net_suite/connection.rb
```diff
@@ -32,6 +25,14 @@ def enabled?
     ENV["CLOUD_ELEMENTS_ORGANIZATION_SECRET"].present?
   end
 
+  def attribute_mapper?
+    true
+  end
+
+  def attribute_mapper
+    super || create_attribute_mapper
```
base10: When I was working with @mike-burns, he recommended against this sort of construction as 1) too implicit and 2) would potentially cause a GET request to have database side-effects. You're balancing concerns differently and I'm curious about your preference here vs. an `after_create` action.

base10:
> I removed gender_map in eaf8227. I'm not seeing any code for job title handling. Can you paste a link?

It looks like you took care of it.

app/models/attribute_mapper.rb
base10: Excellent

app/models/net_suite/connection.rb
jferris: I agree with @mike-burns.

However, this is the pattern that was already set in this application - it's how the `Connection` objects are initialized - and I'd prefer to do an intentional refactor to build the `Connection` family on write vs on read instead of building some elements on write and some on read.

base10: :+1:

base10:
LGTM

app/models/net_suite/export.rb
jferris: Addressed in 2b3b9fe.

jferris:
@base10 ready for re-review.

app/models/net_suite/attribute_mapper.rb
```diff
@@ -16,22 +16,18 @@ def initialize(attribute_mapper:, configuration:)
...
+      exported_profile["title"] = format_job_title(exported_profile["title"])
+    end
   end
 
   private
 
   def format_job_title(value)
     value = Hash(value)
-    value.fetch(:title) { "" }
+    value.fetch("title") { "" }
```
jferris: I don't think this was actually working before, but we didn't have sufficient integration coverage to notice. It may have worked in a browser at some point but become broken during a refactor. We now have integration and unit coverage on post-handling.

base10:
Nice, some very good improvements here. I'll need to make a note about `tap` because that looks very helpful.

jferris:
> I'll need to make a note about `tap` because that looks very helpful.

I actually consider `tap` to be a code smell - it signifies that you're about to mutate the return value - but I prefer it to mutation and explicit return.

base10:
That makes sense. I'll re-up my LGTM.

PR #152, opened 2015-07-29
Update namely gem

macat:
Namely gem version to 0.2.1.https://trello.com/c/6K8Sn80h/37-deprecation-change-from-using-rest-client-to-rest-client


jferris:
The changeset looks good to me. Are there any other changes in the gem we should be worried about?

macat:
Only that rest-client change.

jferris:
Okay, looks good to me, then!

PR #153, opened 2015-07-29
Install specific version of bundler in Dockerfile

jferris:
Because:

* Newer versions of bundler add "BUNDLED_WITH"
* Older versions will remove it

This commit:

* Adds a specific, newer version of bundler
* "BUNDLED_WITH" will no longer change between machines


base10:
LGTM

PR #154, opened 2015-07-29
Include custom Namely fields in possible mappings

jferris:
Because:

* Many NetSuite fields do not have a corresponding field in Namely
* Users will create custom Namely fields to capture these values

This commit:

* Uses the Namely API to discover possible fields
* Includes custom fields in the list of mappable fields

https://trello.com/c/rgC5mN3g


base10:
Looks sharp. I like it. Ship it.

PR #155, opened 2015-07-29
Reset connections on disconnect

jferris:
We store authorization and configuration for connections. Users can
choose to disconnect one of the integrations at any time.

Previously, disconnecting would clear the authorization fields on a
connection. This meant that setting up the connection again would only
ask for authorization, and wouldn't prompt again for configuration.

This commit deletes the connections entirely when disconnecting,
ensuring that all authorization and configuration is removed. The
consequence is that reconnecting will always prompt for everything, even
if we add new fields.

This commit also corrects some association dependencies which were
incorrect. The issues were revealed by deleting the connection objects,
which resulted in deleting the associated user from Namely connect. The
dependencies are now reversed so that connections will delete mappings,
and deleting a user will also delete connections.

https://trello.com/c/5K57ls3e

### DIFF
--- a/app/controllers/connections_controller.rb
+++ b/app/controllers/connections_controller.rb
@@ -12,7 +12,7 @@ class ConnectionsController < IntegrationController
   end
 
   def destroy
-    connection.disconnect
+    connection.destroy
     redirect_to dashboard_path
   end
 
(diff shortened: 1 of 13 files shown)


base10:
LGTM

PR #156, opened 2015-07-29
Fix FieldMapping dependency

jferris:
This dependency caused a stack overflow when disconnecting an
integration where fields had already been mapped, because the field
mapper tried to delete its mappings, and the mappings tried to delete
their mapper.


base10:
LGTM

PR #157, opened 2015-07-29
Create a Profile object, wrap Namely Profiles in Profile

base10:
* Profile will standardize some of how we work with Namely profiles that
  come through the Namely HRIS API
* Profile will handle #title and reach into the :job_title data
  structure to return the underlying title
* #name will return #first_name and #last_name concatenated
* Addresses a bug that manifested in SyncMailer where a name wasn't
  available in a sync result because the profile the result delegated
  the `#name` call to didn't understand `#name`, but as an OpenStruct,
  it returned nil
* For: https://trello.com/c/ANtvl85K/

### DIFF
--- a/app/models/attribute_mapper.rb
+++ b/app/models/attribute_mapper.rb
@@ -29,12 +29,9 @@ class AttributeMapper < ActiveRecord::Base
 
   def export(profile)
     field_mappings.each_with_object({}) do |field_mapping, accumulator|
-      if profile.send(field_mapping.namely_field_name).present?
-        accumulator.merge!(
-          field_mapping.integration_field_name => profile.send(
-            field_mapping.namely_field_name
-          )
-        )
+      value = profile[field_mapping.namely_field_name]
+      if value.present?
+        accumulator.merge!(field_mapping.integration_field_name => value)
       end
     end
   end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -1,0 +1,16 @@
+class Profile
+  delegate :[], to: :namely_profile
+  delegate :update, to: :namely_profile
+
+  def initialize(namely_profile)
+    @namely_profile = namely_profile
+  end
+
+  def name
+    "#{namely_profile[:first_name]} #{namely_profile[:last_name]}"
+  end
+
+  private
+
+  attr_reader :namely_profile
+end
--- a/app/models/net_suite/connection.rb
+++ b/app/models/net_suite/connection.rb
@@ -49,7 +49,7 @@ class NetSuite::Connection < ActiveRecord::Base
   def sync
     NetSuite::Export.new(
       attribute_mapper: net_suite_attribute_mapper,
-      namely_profiles: user.namely_profiles.all,
+      namely_profiles: user.namely_profiles,
       net_suite: client
     ).perform
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -48,7 +48,7 @@ class User < ActiveRecord::Base
   end
 
   def namely_profiles
-    namely_connection.profiles
+    namely_connection.profiles.all.map { |profile| Profile.new(profile) }
   end
 
   def namely_fields
--- a/spec/models/net_suite/attribute_mapper_spec.rb
+++ b/spec/models/net_suite/attribute_mapper_spec.rb
@@ -30,7 +30,7 @@ describe NetSuite::AttributeMapper do
     end
 
     it "doesn't map empty values" do
-      delete_keys = [:email, :last_name]
+      delete_keys = ["email", "last_name"]
       profile_data = stubbed_profile_data
       delete_keys.each { |key| profile_data[key] = nil }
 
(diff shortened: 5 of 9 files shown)


app/models/profile.rb
```diff
@@ -0,0 +1,25 @@
...
+    "#{first_name} #{last_name}"
+  end
+
+  def title
+    job_title["title"]
+  end
+
+  private
+
+  def method_missing(method_name, *arguments, &block)
```
jferris: Are we delegating enough methods that it's worthwhile to use `method_missing`? Some thoughts on why we might avoid it:

* It's easier to know exactly what we're using if we don't
* `method_missing` makes it harder to diagnose bugs
* `method_missing` is harder to understand

app/models/profile.rb
```diff
@@ -0,0 +1,25 @@
...
+    namely_profile.send(method_name, *arguments, &block)
+  end
+
+  def respond_to_missing?(method, *)
+    super
```
jferris: If we're delegating to `namely_profile`, you'll want to use `namely_profile.respond_to?(method) || super`.

https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

app/models/profile.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize(namely_profile)
+    @namely_profile = namely_profile
+  end
+
+  def name
+    "#{first_name} #{last_name}"
+  end
+
+  def title
+    job_title["title"]
```
jferris: If we handle `job_title` here, is there corresponding code we can remove from `NetSuite::AttributeMapper`?

app/models/profile.rb
base10: Good question. I'm in full agreement with `method_missing` being suboptimal.

My thought was that with the Namely custom fields coming through whenever they're added, it'd be difficult to know what to delegate. 

Maybe we could use some meta programming and ask `User#namely_fields`, but that feels harder in a different direction.

I'm very open to alternate suggestions.

app/models/profile.rb
base10: YES! Thought about it and forgot about it. Thanks for the reminder.

app/models/profile.rb
base10: Fixed with 363d5c4

app/models/profile.rb
jferris: Since the fields are mostly being metaprogrammed, what do you think about using `[]` and delegating that to the Namely model? We can intercept fields we want to transform (`title`) and leave others alone.

I think leaving `name` as a top-level method makes sense (and not returning anything for `profile['name']`), because it's not actually a field.

app/models/profile.rb
base10: Fixed in 616d541, but not convinced it's the best way to handle that, particularly in `app/models/profile`

app/models/profile.rb
base10: Giving this a shot.

base10: Actually, after delegating `[]` and removing `method_missing` and `respond_to_missing?` it's apparent I'm unclear about your suggestion. I'll look into this a little further tonight, but if you had a pointer about what might make your suggestion clearer, I'd be interested in seeing it.

jferris: My thinking was that we'd use `[]` everywhere for profile fields (because we don't know if they exist) and use methods (like `name`) for things that we define ourselves.

base10: OK, so, if I'm understanding you correctly, some sample calls would look like:

`profile.name # => "First Last"`
`profile.title # => "Developer"`
`profile["gender"] # => "Female"`

That honestly feels a little strange to me.

jferris: I think the only public methods would be `name` and `[]` (not `title`).

My thinking is that we actually only have two things on a profile:

* `name`: this is what we display in the UI. It has to exist, it has special behavior, it's normalized.
* `[]`: get the text value of a named field

base10: OK, so `title` you'd want to handle with NetSuite::AttributeMapper, then. Alright, let me see how this looks. Thanks.

base10: @jferris This is implemented and ready for rereview.

app/models/attribute_mapper.rb
```diff
@@ -28,17 +28,28 @@ def build_field_mappings(default_field_mapping)
   end
 
   def export(profile)
+    @profile = profile
     field_mappings.each_with_object({}) do |field_mapping, accumulator|
-      if profile.send(field_mapping.namely_field_name).present?
+      if profile.respond_to?(field_mapping.namely_field_name)
```
jferris: What method are we expecting here? My expectation was that we wouldn't need to check for anything but fields, which are handled by `[]`.

app/models/attribute_mapper.rb
```diff
@@ -28,17 +28,28 @@ def build_field_mappings(default_field_mapping)
   end
 
   def export(profile)
+    @profile = profile
```
jferris: Mutating here and then using the value in subsequent methods introduces Connascence of Execution. It means developers have to keep track of order of execution, and future changes will have to worry about disrupting that order. In these cases, it's generally advisable to do one of two things:

* Pass the value as a method argument, avoiding state/mutation.
* Introduce a new class which represents that state and can accept it in the constructor. This means the state is available in all methods for that class regardless of which methods are executed in which order.

app/models/attribute_mapper.rb
```diff
@@ -28,17 +28,28 @@ def build_field_mappings(default_field_mapping)
...
+          field_mapping.integration_field_name => profile[
+            field_mapping.namely_field_name
+          ]
+        )
       end
     end
   end
 
+  def has_value_for_profile_field?(field)
+    !@profile[field].nil? && !@profile[field].empty?
```
jferris: Is this different than `present?`?

app/models/attribute_mapper.rb
base10: I want to catch `#name` here

jferris: My understanding is that, because `name` isn't a field returned by `namely_fields`, users won't ever have the option of mapping something to `name`, so we shouldn't have to catch it here.

base10: Oh, true. Good point.

base10: Fixed with 67c6668

app/models/attribute_mapper.rb
base10: Fixed with 67c6668

app/models/attribute_mapper.rb
base10: Fixed with 67c6668

app/models/attribute_mapper.rb
```diff
@@ -29,11 +29,11 @@ def build_field_mappings(default_field_mapping)
 
   def export(profile)
     field_mappings.each_with_object({}) do |field_mapping, accumulator|
-      if profile.send(field_mapping.namely_field_name).present?
+      if profile[field_mapping.namely_field_name].present?
```
jferris: Thoughts on extracting an intention-revealing variable like `value` from `profile[field_mapping.namely_field_name]`?

app/models/profile.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def initialize(namely_profile)
+    @namely_profile = namely_profile
+  end
+
+  def name
+    "#{namely_profile[:first_name]} #{namely_profile[:last_name]}"
+  end
+
+  attr_reader :namely_profile
```
jferris: Can this be private?

jferris:
I had a couple of minor comments. Good to squash and merge at your discretion.

app/models/attribute_mapper.rb
base10: That's a nice change. Thanks for the suggestion.

app/models/profile.rb
base10: Yep. I think that was inadvertently deleted when I cut the title handling. Good catch.

PR #158, opened 2015-07-30
Extract generic bulk sync class

jferris:
Because we're doing bulk import and bulk export for several
integrations, and the `NetSuite::BulkExport` class had very little
NetSuite-specific behavior, this commit renames and generalizes that
class to work for all integrations.

This allows us to remove the specialized `Jobvite::BulkImport` class,
reducing the amount of code we need to adjust as we introduce field
mapping.


base10:
Nice!

LGTM

PR #159, opened 2015-07-30
Users can control field mappings for Jobvite

jferris:
* Uses default mappings for existing connections
* Presents the field mapping form for new connections
* Also allows date fields to be mapped

https://trello.com/c/y10H2LkJ

### DIFF
--- a/spec/models/jobvite/connection_spec.rb
+++ b/spec/models/jobvite/connection_spec.rb
@@ -2,6 +2,7 @@ require "rails_helper"
 
 describe Jobvite::Connection do
   describe "associations" do
+    it { should belong_to(:attribute_mapper).dependent(:destroy) }
     it { should belong_to(:user) }
   end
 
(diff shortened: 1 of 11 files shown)


spec/models/jobvite/connection_spec.rb
```diff
@@ -42,10 +43,35 @@
...
+  describe "#attribute_mapper" do
+    it "builds and saves an attribute mapper" do
+      connection = build(:jobvite_connection)
+
+      connection.save!
+
+      expect(connection.attribute_mapper).to be_an_instance_of(AttributeMapper)
+      expect(connection.attribute_mapper).to be_persisted
+      expect(mapped_fields_for(connection.attribute_mapper)).
+        to match_array([
```
base10: Nice way of handling this.

base10:
Very nice. LGTM.

PR #160, opened 2015-07-30
Refactor attribute mapper initialization

jferris:
Because each integration needs to initialize an attribute mapper with
defaults, it makes sense to consolidate the logic for doing so.

This commit replaces the NetSuite::AttributeMapperBuilder class with a
generalized version usable by any integration's connection class.


base10:
👏 Love it. LGTM

PR #161, opened 2015-07-30
Use different words for mapping/normalization

jferris:
It's confusing to name both the high-level mapper model and the
integration-specific normalization strategies "attribute mappers."

This commit switches to consistently calling the attribute mapper model
"attribute mapper," and calling the strategies "normalizer."


base10:
Nice clean-up. LGTM.

PR #162, opened 2015-08-04
Contact NetSuite, get a list of employees, determine profile fields

base10:
* NetSuite, through Cloud Elements, doesn't have a way to directly list
  what fields are available on profiles
* We want to get a full list of available fields for each integration
* Introduces NetSuite::EmployeeField
* Created fro NetSuite::Client#profile_fields
* We aren't storing the information, but making it available when field
  mappings between Namely and NetSuite are edited
* This does not yet handle custom fields, since those are going to be
  more complex
* Since we don't want to count on Employee ids within NetSuite, we
  request 5 employees, since that's the smallest window allowed for
  getting a list of employees
* Matches, where it can, the types of fields that we support in Namely
  for mapping
  * date
  * text
  * email
* Also identifies other types that we don't yet handle in Namely for
  mapping
  * boolean
  * fixnum
  * object
* The names for the non-supported field types may change
* NetSuite::Client#profile_fields just hands back a list of
  NetSuite::EmployeeField objects and presumes supported type filtering
  is handled downstream
* For: https://trello.com/c/mThbJz4M/

### DIFF
--- a/app/models/net_suite/client.rb
+++ b/app/models/net_suite/client.rb
@@ -1,5 +1,10 @@
 module NetSuite
   class Client
+    REQUEST_BASE = "/hubs/erp"
+    EMPLOYEE_REQUEST = REQUEST_BASE + "/employees"
+    SUBSIDIARY_REQUEST = REQUEST_BASE + "/lookups/subsidiary"
+    INSTANCES = "/instances"
+
     delegate :get_json, to: :request
     delegate :submit_json, to: :request
 
@@ -35,7 +40,7 @@ module NetSuite
     def create_instance(params)
       submit_json(
         :post,
-        "/instances",
+        INSTANCES,
         "configuration" => {
           "user.username" => params[:email],
           "user.password" => params[:password],
@@ -53,7 +58,7 @@ module NetSuite
     def create_employee(params)
       submit_json(
         :post,
-        "/hubs/erp/employees",
+        EMPLOYEE_REQUEST,
         params
       )
     end
@@ -61,13 +66,19 @@ module NetSuite
     def update_employee(id, params)
       submit_json(
         :patch,
-        "/hubs/erp/employees/#{id}",
+        "#{EMPLOYEE_REQUEST}/#{id}",
         params
       )
     end
 
     def subsidiaries
-      get_json("/hubs/erp/lookups/subsidiary")
+      get_json(SUBSIDIARY_REQUEST)
+    end
+
+    def profile_fields
+      @profile_fields ||= NetSuite::EmployeeFieldsLoader.new(
+        request: request
+      ).load_profile_fields
     end
 
     def request
--- a/app/models/net_suite/employee_field.rb
+++ b/app/models/net_suite/employee_field.rb
@@ -1,0 +1,18 @@
+module NetSuite
+  class EmployeeField
+    attr_reader :name, :value
+
+    def initialize(name:, value:)
+      @name = name
+      @value = value
+    end
+
+    def label
+      name.titleize
+    end
+
+    def type
+      TypeForField.for_field(name: name, value: value)
+    end
+  end
+end
--- a/spec/models/net_suite/client_spec.rb
+++ b/spec/models/net_suite/client_spec.rb
@@ -262,17 +262,41 @@ describe NetSuite::Client do
         ).
         to_return(status: 200, body: subsidiaries.to_json)
 
-      client = NetSuite::Client.new(
-        user: build_stubbed(:user),
-        user_secret: "user-secret",
-        organization_secret: "org-secret",
-        element_secret: "element-secret"
-      )
-
       result = client.subsidiaries
 
       expect(result).to be_success
       expect(result.to_a).to eq(subsidiaries)
     end
+  end
+
+  describe "#profile_fields" do
+    it "gets a currest list of NetSuite employee profile fields" do
+      fields = [
+        double(:employee_field),
+        double(:employee_field)
+      ]
+
+      fields_loader = instance_spy(
+        NetSuite::EmployeeFieldsLoader,
+        load_profile_fields: fields
+      )
+
+      netsuite_client = client
+
+      allow(NetSuite::EmployeeFieldsLoader).to receive(:new).
+        with(request: netsuite_client.request).
+        and_return(fields_loader)
+
+      expect(netsuite_client.profile_fields).to match_array(fields)
+    end
+  end
+
+  def client
+    NetSuite::Client.new(
+      user: build_stubbed(:user),
+      user_secret: "user-secret",
+      organization_secret: "org-secret",
+      element_secret: "element-secret"
+    )
   end
 end
--- a/spec/models/net_suite/employee_field_spec.rb
+++ b/spec/models/net_suite/employee_field_spec.rb
@@ -1,0 +1,37 @@
+require "rails_helper"
+
+describe NetSuite::EmployeeField do
+  describe "#label" do
+    it "provides a valid label for camelCase" do
+      employee_field = described_class.new(name: "homePhone", value: "foo")
+
+      expect(employee_field.label).to eq("Home Phone")
+    end
+
+    it "provides a valid label for doubleCamelCase" do
+      employee_field = described_class.new(
+        name: "globalSubscriptionStatus",
+        value: {}
+      )
+
+      expect(employee_field.label).to eq("Global Subscription Status")
+    end
+
+    it "provides a valid label for single words" do
+      employee_field = described_class.new(name: "phone", value: "Foo")
+
+      expect(employee_field.label).to eq("Phone")
+    end
+  end
+
+  describe "#type" do
+    it "provides a type" do
+      employee_field = described_class.new(
+        name: "email",
+        value: "test@example.com"
+      )
+
+      expect(employee_field.type).to eq("email")
+    end
+  end
+end
--- a/app/models/net_suite/employee_fields_loader.rb
+++ b/app/models/net_suite/employee_fields_loader.rb
@@ -1,0 +1,27 @@
+module NetSuite
+  class EmployeeFieldsLoader
+    EMPLOYEE_LIMIT = 5
+
+    def initialize(request:)
+      @request = request
+    end
+
+    def load_profile_fields
+      employees = request.get_json(
+        "#{NetSuite::Client::EMPLOYEE_REQUEST}?pageSize=#{EMPLOYEE_LIMIT}"
+      )
+
+      convert_employee_data_to_profile_fields(employees.first)
+    end
+
+    def convert_employee_data_to_profile_fields(employee)
+      employee.map do |name, value|
+        NetSuite::EmployeeField.new(name: name, value: value)
+      end
+    end
+
+    private
+
+    attr_reader :request
+  end
+end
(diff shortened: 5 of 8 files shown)


base10:
@drapergeek && @tabfugnic, Would appreciate your feedback

app/models/net_suite/client.rb
```diff
@@ -121,6 +125,17 @@ def secrets
       }
     end
 
+    def load_profile_fields
+      employees = get_json("/hubs/erp/employees?pageSize=5")
```
drapergeek: What do you think about extracting the URL and the page size here to constants?

app/models/net_suite/client.rb
```diff
@@ -121,6 +125,17 @@ def secrets
...
+      employee_to_profile_fields(employees.first)
+    end
+
+    def employee_to_profile_fields(employee)
```
drapergeek: Is this a method name that is used elsewhere so you're trying to match it? If not, is there a more descriptive name we could use here?

app/models/net_suite/client.rb
```diff
@@ -121,6 +125,17 @@ def secrets
       }
     end
 
+    def load_profile_fields
```
drapergeek: This class is getting pretty big (already over the 100 mark before the private area and before this new addition).

What do you think about letting this method call out to another object to handle loading the profile fields? Seems like that would be a clean place to separate the responsibilities.

You could also inject that object so that in testing it is easier to skip that step when you want.

app/models/net_suite/client.rb
```diff
@@ -69,6 +69,10 @@ def subsidiaries
       get_json("/hubs/erp/lookups/subsidiary")
     end
 
+    def profile_fields
```
drapergeek: This class is getting pretty big (already over the 100 mark before the private area and before this new addition).

What do you think about letting this method call out to another object to handle loading the profile fields? Seems like that would be a clean place to separate the responsibilities.

You could also inject that object so that in testing it is easier to skip that step when you want.

app/models/net_suite/employee_field.rb
```diff
@@ -0,0 +1,51 @@
+module NetSuite
+  class EmployeeField
+    attr_accessor :name, :value
```
drapergeek: Does this need to be an accessor? Could we use the reader instead?

spec/fixtures/api_responses/net_suite_employees.json
```diff
@@ -0,0 +1,366 @@
+[
```
drapergeek: What do you think about limiting the information in here? From the looks of your spec, you don't seem to use a large majority of this file, can we remove parts of it?

Also what do you think about converting this to be a ruby method so that you can specify things that you're going to use in your tests? By doing that, whoever is reading/modifying the tests doesn't necessarily have to look at this file to understand what is going on.

spec/models/net_suite/employee_field_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    it "provides a valid label for single words" do
+      employee_field = described_class.new(name: "phone", value: "Foo")
+
+      expect(employee_field.label).to eq("Phone")
+    end
+  end
+
+  describe "#type" do
```
drapergeek: It seems that all of these are actually testing the underlying checks for the type. What do you think about extracting the object that handles that into its own class? That seems like something you could use in another implementations and it would make it more clear what is being tested.

app/models/net_suite/client.rb
base10: Addressed with bc3c3ec

app/models/net_suite/client.rb
base10: Addressed with d1a3c98

app/models/net_suite/employee_field.rb
base10: Fixed with 8cc5e7a

spec/fixtures/api_responses/net_suite_employees.json
base10: I'm fine with trimming with what's in the file. 

If I follow the second recommendation, is that in place of using the fixture file altogether or treating it like we would a factory where you get default info, but to avoid Mystery Guests, you specify what you're expecting to see?

drapergeek: >treating it like we would a factory where you get default info, but to avoid Mystery Guests, you specify what you're expecting to see?

That one.

spec/models/net_suite/employee_field_spec.rb
base10: I'm not sure I follow your suggestion. Are you suggesting extracting `#type` and the underlying methods to their own class that would receive the NetSuite::EmployeeField object and answer questions about it?

Alternately, thoughts on making the `?` methods public and testing those directly?

drapergeek: I am suggesting an object that will return the type for a given variable.

```ruby
def type
  TypeForField.for_field(name)
end
```

Let that object handle all the questions of `boolean` and such. That way the test is much closer to what you actually care about.

base10: OK, addressed with 37cb9fc. Please let me know if you were thinking about something different than that.

drapergeek:
@base10 seems like this one got closed in a diff: https://github.com/namely/connect/pull/162#discussion_r36211327

base10:
@drapergeek Yep. Looking at that now and thinking through how I want to handle it.

app/models/net_suite/client.rb
base10: Looking at the class and what all needs to happen with it, I extracted NetSuite::Client::Request and NetSuite::Client::Result in bb8ad88. I'm debating pulling out another class for doing the heavy lifting on loading the employee fields.

base10: See also 1715178 and 978ba504

base10: I moved the extraction of NetSuite::Client::Request and Result into #163. to keep this PR focused, per feedback. NetSuite::EmployeeField is still extracted as its own class. Visible as of da49739.

spec/models/net_suite/client_spec.rb
```diff
@@ -262,17 +262,40 @@
...
+  describe "#profile_fields" do
+    it "gets a currest list of NetSuite employee profile fields" do
+      client = stubbed_client
+
+      fields = [
+        NetSuite::EmployeeField.new(name: "firstName", value: "Ralph"),
+        NetSuite::EmployeeField.new(name: "lastName", value: "Bot")
+      ]
+
+      allow(NetSuite::EmployeeField).to receive(
```
drapergeek: This is stubbing the system under test (SUT). If you extract the class that is being used in this case into it's own object then you could instead pass in a fake here and avoid the stubbing altogether.

spec/models/net_suite/employee_field_spec.rb
```diff
@@ -0,0 +1,75 @@
+require "rails_helper"
+
+describe NetSuite::EmployeeField do
+  describe ".retrieve_fields" do
```
drapergeek: The majority of the test has to do with the private class inside the `EmployeeField` class. What do you think about extracting that class into a first class object so that this test can be more focused.

base10: I'll try that later this evening. 

My thinking was that I wouldn't be using that private class outside of case, so it made sense to define it as a private class, particularly since I didn't care to expose the methods of `RequestHelper`.

spec/models/net_suite/client_spec.rb
base10: I can try that. I am curious why this falls under "stubbing the system under test" but passing in a fake wouldn't be, though.

spec/models/net_suite/employee_field_spec.rb
drapergeek: >My thinking was that I wouldn't be using that private class outside of case, so it made sense to define it as a private class

That is not always the best metric. In this case the question is, "when this spec fails, how far away is the actual problem?"

In this case I'd extract it just to test closer to the object that we're working with.

spec/models/net_suite/client_spec.rb
drapergeek: In your case, you are actively calling "stub" on the object you're testing which means that you have internal knowledge about what the method is doing. If in the future, that method changes, this test may be lying to you.

By pulling out the dependency, you no longer know what method even calls this dependency, only that you pass in an object to handle the client portion of this. You can pass in your normal client, a fake client or even a client that handles XML in the future and you don't have to worry about knowing the internal workings.

Great post that explains things better: https://robots.thoughtbot.com/don-t-stub-the-system-under-test

base10: OK, so, let me make sure I understand:

I have `NetSuite::Client` and a method on it to get profile fields. Because that method calls `NetSuite::EmployeeFields`, `NetSuite::EmployeeFields` is then under test as well as `NetSuite::Client` and should't be stubbed?

I've understood not stubbing the system under test to mean not stubbing out parts of your own object at the unit testing level. For instance, I'd expect that stubbing a method on `NetSuite::Client` when I'm testing `NetSuite::Client` would be a problem. I had not understood that at the unit level as crossing over to, say, stubbing `NetSuite::FooBar`.

spec/fixtures/api_responses/net_suite_employees.json
base10: Addressed with 64ea82d. I expect when we resolve questions about the class structure, this will move again.

spec/models/net_suite/client_spec.rb
drapergeek: I apologize, I misread the difference between these two objects. I thought this was the `Fields` test. Going back and forth between specs and code I got them confused.

That being said, I would still recommend pulling the `RequestHelper` out to a first class object. That code is very unrelated to the other code in the class.

base10: :+1:

base10:
@drapergeek Classes extracted as requested and the interface to determining test data is improved as discussed. Ready for rereview.

spec/models/net_suite/client_spec.rb
```diff
@@ -262,17 +262,40 @@
...
       result = client.subsidiaries
 
       expect(result).to be_success
       expect(result.to_a).to eq(subsidiaries)
     end
   end
+
+  describe "#profile_fields" do
+    it "gets a currest list of NetSuite employee profile fields" do
+      client = stubbed_client
```
drapergeek: This client isn't actually being stubbed out so it's a bit confusing to follow. Can this be named something else?

spec/models/net_suite/client_spec.rb
drapergeek: Is this stubbing still necessary after the refactoring?

spec/models/net_suite/client_spec.rb
base10: Fixed with 4dd5ac2

spec/models/net_suite/client_spec.rb
base10: I'm now stubbing the correct class. See 39fd3c2.

drapergeek:
LGTM

PR #163, opened 2015-08-05
Extract Request and Result classes from NetSuite::Client

base10:
NetSuite::Client is getting a bit large and with retrieving fields
from NetSuite and having to build yet another object, it makes sense to
extract these now.

* In support of PR #162
* For: https://trello.com/c/mThbJz4M/

### DIFF
--- a/app/models/net_suite/client.rb
+++ b/app/models/net_suite/client.rb
@@ -1,6 +1,7 @@
 module NetSuite
   class Client
-    BASE_URL = "https://api.cloud-elements.com/elements/api-v2"
+    delegate :get_json, to: :request
+    delegate :submit_json, to: :request
 
     def self.from_env(user)
       new(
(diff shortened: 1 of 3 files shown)


base10:
@drapergeek Per discussion, extracted this outside the context of #162. Tried flipping the order on this and seemed like it was going to work better to get this out of the way first. Review appreciated.

app/models/net_suite/client.rb
```diff
@@ -69,85 +70,12 @@ def subsidiaries
...
-        json.each(&block)
-      end
-
-      private
-
-      def json
-        @json ||= JSON.parse(@response)
-      end
+    def request
+      @request || Request.new(
```
drapergeek: `||=` maybe?

base10: :disappointed: Fixed in 4498b7d

drapergeek:
LGTM

base10:
\o/ Thanks.

PR #164, opened 2015-08-11
Introduce installations

jferris:
Because:

* Multiple users belong to the same Namely account
* Only one connection can sync for each integration/Namely account

This commit:

* Introduces an invisible installation concept
* Adds users from the same Namely account to the same installation
* Associates connections with installations instead of users

Effects:

* Multiple users on the same Namely account can share connections

https://trello.com/c/W6KqJcez

### DIFF
--- a/app/jobs/sync_job.rb
+++ b/app/jobs/sync_job.rb
@@ -1,7 +1,7 @@
 class SyncJob
-  def initialize(integration_id, user_id)
+  def initialize(integration_id, installation_id)
     @integration_id = integration_id
-    @user_id = user_id
+    @installation_id = installation_id
   end
 
   def perform
@@ -14,26 +14,28 @@ class SyncJob
   private
 
   def connection
-    user.send(:"#{@integration_id}_connection")
+    installation.public_send("#{integration_id}_connection")
   end
 
   def deliver_sync_notification(results)
-    SyncMailer.sync_notification(
-      email: user.email,
-      integration_id: integration_id,
-      results: results
-    ).deliver
+    installation.users.each do |user|
+      SyncMailer.sync_notification(
+        email: user.email,
+        integration_id: integration_id,
+        results: results
+      ).deliver
+    end
   end
 
   def notifier
     AuthenticationNotifier.new(
       integration_id: integration_id,
-      user: user
+      installation: installation
     )
   end
 
-  def user
-    @user ||= User.find(@user_id)
+  def installation
+    @installation ||= Installation.find(@installation_id)
   end
 
   attr_reader :integration_id
--- a/app/models/installation.rb
+++ b/app/models/installation.rb
@@ -1,0 +1,67 @@
+class Installation < ActiveRecord::Base
+  has_many :users, dependent: :destroy
+  has_one(
+    :greenhouse_connection,
+    class_name: "Greenhouse::Connection",
+    dependent: :destroy
+  )
+  has_one(
+    :icims_connection,
+    class_name: "Icims::Connection",
+    dependent: :destroy
+  )
+  has_one(
+    :jobvite_connection,
+    class_name: "Jobvite::Connection",
+    dependent: :destroy
+  )
+  has_one(
+    :net_suite_connection,
+    class_name: "NetSuite::Connection",
+    dependent: :destroy
+  )
+
+  validates :subdomain, presence: true, uniqueness: true
+
+  def self.ready_to_sync_with(integration)
+    association = "#{integration}_connection"
+    joins(association.to_sym).
+      where(association.pluralize => { found_namely_field: true })
+  end
+
+  def jobvite_connection
+    super || Jobvite::Connection.create!(installation: self)
+  end
+
+  def icims_connection
+    super || Icims::Connection.create!(installation: self)
+  end
+
+  def greenhouse_connection
+    super || Greenhouse::Connection.create!(installation: self)
+  end
+
+  def net_suite_connection
+    super || NetSuite::Connection.create!(installation: self)
+  end
+
+  def send_connection_notification(*arguments)
+    users.each do |user|
+      user.send_connection_notification(*arguments)
+    end
+  end
+
+  def namely_connection
+    owner.namely_connection
+  end
+
+  def namely_profiles
+    owner.namely_profiles
+  end
+
+  private
+
+  def owner
+    users.first
+  end
+end
--- a/app/services/candidate_importer.rb
+++ b/app/services/candidate_importer.rb
@@ -32,38 +32,42 @@ class CandidateImporter
   def namely_importer
     NamelyImporter.new(
       normalizer: import_assistant.normalizer,
-      namely_connection: user.namely_connection
+      namely_connection: installation.namely_connection
     )
   end
 
   def notifier
     AuthenticationNotifier.new(
       integration_id: assistant_class::INTEGRATION_ID,
-      user: user
+      installation: installation
     )
   end
 
-  def user
-    connection.user
+  def installation
+    connection.installation
   end
 
   private
 
   def notify_of_successful_import
-    mailer.delay.successful_import(
-      candidate: import_assistant.candidate,
-      email: user.email,
-      integration_id: assistant_class::INTEGRATION_ID
-    )
+    installation.users.each do |user|
+      mailer.delay.successful_import(
+        candidate: import_assistant.candidate,
+        email: user.email,
+        integration_id: assistant_class::INTEGRATION_ID
+      )
+    end
   end
 
   def notify_of_unsuccessful_import
-    mailer.delay.unsuccessful_import(
-      candidate: import_assistant.candidate,
-      email: user.email,
-      integration_id: assistant_class::INTEGRATION_ID,
-      status: import_assistant.import_candidate
-    )
+    installation.users.each do |user|
+      mailer.delay.unsuccessful_import(
+        candidate: import_assistant.candidate,
+        email: user.email,
+        integration_id: assistant_class::INTEGRATION_ID,
+        status: import_assistant.import_candidate
+      )
+    end
   end
 
   def report_import_results
--- a/db/migrate/20150807150624_adjust_user_add_reference_to_installation.rb
+++ b/db/migrate/20150807150624_adjust_user_add_reference_to_installation.rb
@@ -1,0 +1,19 @@
+class AdjustUserAddReferenceToInstallation < ActiveRecord::Migration
+  def up
+    add_reference :users, :installation, index: true, foreign_key: true
+
+    update(<<-SQL)
+      UPDATE users SET installation_id = (
+        SELECT installations.id
+        FROM installations
+        WHERE installations.subdomain = users.subdomain
+      )
+    SQL
+
+    change_column_null :users, :installation_id, false
+  end
+
+  def down
+    remove_reference :users, :installation
+  end
+end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,6 +1,5 @@
 FactoryGirl.define do
   factory :attribute_mapper do
-    user
   end
 
   factory :field_mapping do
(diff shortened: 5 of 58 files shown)


app/jobs/sync_job.rb
```diff
@@ -14,26 +14,28 @@ def perform
   private
 
   def connection
-    user.send(:"#{@integration_id}_connection")
+    installation.send(:"#{@integration_id}_connection")
```
sikachu: Do you think it's necessary to use `#send`? Or, will it work with `#public_send`?

jferris: It works with `public_send`. Addressed in 843cec9.

kenyonj: I don't think you need the preceeding `:` there... Does that do anything different than `.send("#{@integration_id}_connection")` ?

app/jobs/sync_job.rb
```diff
@@ -14,26 +14,28 @@ def perform
   private
 
   def connection
-    user.send(:"#{@integration_id}_connection")
+    installation.public_send(:"#{@integration_id}_connection")
```
kenyonj: I don't think you need the preceeding `:` there. Does that do anything different than `public_send("#{@integration_id}_connection")` ?

app/models/installation.rb
```diff
@@ -0,0 +1,67 @@
...
+  validates :subdomain, presence: true, uniqueness: true
+
+  def self.ready_to_sync_with(integration)
+    association = "#{integration}_connection"
+    joins(association.to_sym).
+      where(association.pluralize => { found_namely_field: true })
+  end
+
+  def jobvite_connection
+    super || Jobvite::Connection.create(installation: self)
```
sikachu: Do you think this would be better as `create!`? I feel like you could ended up with an unpersisted record here.

app/jobs/sync_job.rb
```diff
@@ -14,26 +14,28 @@ def perform
...
       integration_id: integration_id,
-      user: user
+      installation: installation
     )
   end
 
-  def user
-    @user ||= User.find(@user_id)
+  def installation
+    @installation ||= Installation.find(@installation_id)
```
kenyonj: thoughts on moving this over to `@_installation` to designate that you won't be using it directly?

app/jobs/sync_job.rb
jferris: It will work with or without the `:`. Using the `:` will generate a symbol, whereas not using it will generate a string.

In general, using symbols for a closed set of literals is more efficient, because additional strings won't be constructed and comparison can be performed by identity (because symbols are immutable).

In this case, we're dynamically creating a symbol (although the set is still closed). I'm unsure of the performance characteristics for dynamic internalization.

Either way, I don't believe this code path is performance-critical enough that we need to worry about string allocation. Because building a string is slightly simpler, I removed the symbol in 6066095.

app/jobs/sync_job.rb
jferris: We haven't been following that convention in this project, so I don't believe code authors will understand that designation.

app/models/installation.rb
jferris: Yes, I think `create!` is likely better, since we don't handle the failure case.

This code was extracted from `User`, but I'll see if it breaks anything to switch to `create!`.

db/migrate/20150807150624_adjust_user_add_reference_to_installation.rb
```diff
@@ -0,0 +1,19 @@
+class AdjustUserAddReferenceToInstallation < ActiveRecord::Migration
+  def up
+    add_reference :users, :installation, index: true, foreign_key: true
+
+    update(<<-SQL)
+      UPDATE users SET installation_id = (
```
sikachu: How many users are we going to run this migration against? This subquery could take quite some time to run.

app/models/installation.rb
jferris: The tests pass and the UI continues to respond using `create!`, so I introduced it in 035782c.

db/migrate/20150807150624_adjust_user_add_reference_to_installation.rb
jferris: The number of users is fairly low, and there are indexes available for the search part of this query.

spec/factories.rb
```diff
@@ -75,5 +90,10 @@
     sequence(:refresh_token) { |n| "refresh-token-#{n}" }
     access_token_expiry { 15.minutes.from_now }
     email "integrationlover@example.com"
+    installation { Installation.find_or_create_by!(subdomain: subdomain) }
```
sikachu: Wouldn't this make it create a new subdomain record even if you do `build_stubbed`?

app/jobs/sync_job.rb
```diff
@@ -14,26 +14,28 @@ def perform
   private
 
   def connection
-    user.send(:"#{@integration_id}_connection")
+    installation.public_send("#{@integration_id}_connection")
```
seanpdoyle: Would a private `attr_reader` make sense for `@integration_id`?

app/services/candidate_importer.rb
```diff
@@ -32,38 +32,42 @@ def import_assistant
...
       integration_id: assistant_class::INTEGRATION_ID,
-      user: user
+      installation: installation
     )
   end
 
-  def user
-    connection.user
+  def installation
+    connection.installation
```
kenyonj: It looks like `installation` is being used multiple times, should this be memoized?

spec/models/installation_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+        :installation,
+        subdomain: "ready-with-given-service"
+      )
+      create(
+        :net_suite_connection,
+        :connected,
+        :with_namely_field,
+        installation: installation
+      )
+      installation = create(:installation, subdomain: "connected-without-field")
```
sikachu: These setups are a bit hard to follow. I feel like it'd be better put in a well named method instead, but I'll let you decide.

app/jobs/sync_job.rb
seanpdoyle: Looks like that's already the case on line `41`.

Could we use that here?

spec/models/authentication_notifier_spec.rb
```diff
@@ -16,28 +16,29 @@
...
-        user: user_spy
+        installation: installation_spy
       )
       exception = Unauthorized.new(Unauthorized::DEFAULT_MESSAGE)
 
       expect(Rails.logger).to receive(:error).with(
         "#{exception.class} error #{exception.message} for " \
-        "user_id: #{user_spy.id} with #{notifier.integration_name}"
+        "installation_id: #{installation_spy.id} " \
+        "with #{notifier.integration_name}"
```
seanpdoyle: Would it make sense to match against a regex, in case minor changes are made to the error message?

spec/models/installation_spec.rb
```diff
@@ -0,0 +1,157 @@
+require "rails_helper"
+
+describe Installation do
+  describe "validations" do
+    subject { create(:installation) }
+    it { should validate_presence_of(:subdomain) }
```
seanpdoyle: We've used `is_expected.to` elsewhere in the codebase.

Would it make sense to standardize to one particular style?

spec/models/installation_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+      installation = create(
+        :installation,
+        subdomain: "ready-with-different-service"
+      )
+      create(
+        :greenhouse_connection,
+        :connected,
+        :with_namely_field,
+        installation: installation
+      )
```
seanpdoyle: Could these unassigned calls to FactoryGirl be wrapped up in a helper that describes how they're associated?

spec/models/installation_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+        expect(net_suite_connection).to be_a NetSuite::Connection
+        expect(net_suite_connection).to be_persisted
+        expect(net_suite_connection.installation_id).to eq installation.id
+      end
+    end
+  end
+
+  describe "#send_connection_notification" do
+    it "delegates to each users" do
+      users = [build_stubbed(:user), build_stubbed(:user)]
```
seanpdoyle: We're using this exact line in multiple specs.

What do you think about extracting a `build_users` helper?

If the quantity of users built matters (it doesn't seem to), the helper could accept a count.

seanpdoyle: Alternatively, we could use a `build_stubbed_list(2, :user)` (if such a thing exists)

spec/factories.rb
jferris: Yes; these factory changes will perform unnecessary database I/O.

I did this in the interest of trying to minimize changes in the specs, but I'll take a look at removing the `create` calls introduced here.

spec/models/user_spec.rb
```diff
@@ -141,6 +55,65 @@
...
+      expect(Namely::Connection).to have_received(:new).with(
+        access_token: "MY_ACCESS_TOKEN",
+        subdomain: "ellingsonmineral",
+      )
+    end
+  end
+
+  describe "#namely_fields_by_label" do
+    it "returns mappable fields from a Namely connection" do
+      ["single_select", "short_text", "long_text", "number"]
```
seanpdoyle: What is going on with this line?

It looks like an array that isn't assigned, and is inaccessible elsewhere.

spec/factories.rb
jferris: It introduced some noise, but it's not too bad. I removed the `create` calls from factory definitions in 8dc17a2.

app/services/candidate_importer.rb
jferris: Can you explain the advantage you see to memoizing here? As I see it:

* Memoization prevents mutations from being lost (we aren't mutating)
* Memoization prevents recalculations (associations are cached by ActiveRecord)
* Memoization is slightly more complex
* Memoization is less readable
* Memoization can introduce bugs

In general, I avoid memoization unless I know there's a performance problem I'm addressing.

spec/models/installation_spec.rb
jferris: Yeah, I extracted this from `user_spec.rb`, but it looks pretty weird now that I read it more closely. I'll take a look at it.

jferris: Addressed in 58b08e4.

app/jobs/sync_job.rb
jferris: Done in ab0d547.

spec/models/installation_spec.rb
jferris: Done in 58b08e470160c5a5ddb21c7d668887c7921835a4.

spec/models/authentication_notifier_spec.rb
jferris: I'm not sure it's worthwhile. There are enough dynamic portions in the string that order changes will cause failures, and I think the string version is more readable than a regular expression.

spec/models/installation_spec.rb
jferris: Yeah, I went for local consistency, but it is a little weird to see both in the same diff.

I don't want to rewrite every use of `should`, as that's not what's really going on in this changeset, but I switched all uses affected by these changes to use `is_expected.to` in 74c218d.

app/services/candidate_importer.rb
kenyonj: ok, I wasn't sure what `installation` was doing on the `connection` but if it's not calculating anything or re-pulling something from the db then I agree.

spec/models/user_spec.rb
jferris: This was extracted from `attribute_mapper_spec.rb`. As you pointed out, it doesn't do anything. Removed in 4cda597.

spec/models/installation_spec.rb
jferris: I believe this is incidental duplication, rather than duplicated logic. I find the duplicate call to be clearer than the `*_list` versions.

jferris:
@kenyonj @sikachu @seanpdoyle ready for re-review.

seanpdoyle:
LGTM

PR #165, opened 2015-08-17
Users can map custom NetSuite fields

jferris:
Because:

* Users can create custom fields in NetSuite
* Users can create custom fields in Namely

This commit:

* Downloads an employee as an example
* Uses the employee to determine available fields
* Allows users to map available text fields in NetSuite

https://trello.com/c/Z2SKrJeU

### DIFF
--- a/app/models/attribute_mapper_factory.rb
+++ b/app/models/attribute_mapper_factory.rb
@@ -4,7 +4,7 @@ class AttributeMapperFactory
     @connection = connection
   end
 
-  def build_with_defaults(defaults)
+  def build_with_defaults(&defaults)
     @attribute_mapper || assign_attribute_mapper(defaults)
   end
 
@@ -13,7 +13,7 @@ class AttributeMapperFactory
   def assign_attribute_mapper(defaults)
     AttributeMapper.create!.tap do |attribute_mapper|
       @connection.update!(attribute_mapper: attribute_mapper)
-      defaults.each do |integration_field_name, namely_field_name|
+      defaults.call.each do |integration_field_name, namely_field_name|
         attribute_mapper.field_mappings.create!(
           integration_field_name: integration_field_name,
           namely_field_name: namely_field_name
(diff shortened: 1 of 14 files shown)


app/models/attribute_mapper_factory.rb
```diff
@@ -13,7 +13,7 @@ def build_with_defaults(defaults)
   def assign_attribute_mapper(defaults)
     AttributeMapper.create!.tap do |attribute_mapper|
       @connection.update!(attribute_mapper: attribute_mapper)
-      defaults.each do |integration_field_name, namely_field_name|
+      defaults.call.each do |integration_field_name, namely_field_name|
```
JoelQ: Why are defaults callable objects now instead of a hash? What prompted the decision to change?

JoelQ:
One question, otherwise looks good :+1:

app/models/jobvite/connection.rb
```diff
@@ -27,13 +27,15 @@ def attribute_mapper?
...
-        build_with_defaults(
-          "first_name" => "first_name",
-          "last_name" => "last_name",
-          "email" => "email",
-          "start_date" => "start_date",
-          "gender" => "gender",
-        )
+        build_with_defaults do
+          {
+            "first_name" => "first_name",
```
drapergeek: Does order matter? Would moving this to alphabetical make it easier to what's available or missing in future changes?

app/models/attribute_mapper_factory.rb
jferris: For the NetSuite integration, we need to perform an API call to determine the available fields in order to calculate defaults. We don't want to perform that call every time, so we provide the defaults procedure as a block and allow the factory to evaluate it only if necessary.

app/models/jobvite/connection.rb
jferris: Ah, I missed this comment before merging.

I think alphabetizing would be nice, but isn't a big deal.

PR #167, opened 2015-08-18
Refactor API for field mappings

jferris:
We need to add more than just to/from, because we need to store
attribute metadata from integration responses.

This commit expands the API from a hash to method calls so that several
fields can be provided.

### DIFF
--- a/app/models/field_mapping.rb
+++ b/app/models/field_mapping.rb
@@ -5,6 +5,12 @@ class FieldMapping < ActiveRecord::Base
   validates :attribute_mapper_id, presence: true
   validates :integration_field_name, presence: true
 
+  def self.map!(field, to: nil)
+    if where(integration_field_name: field).empty?
+      create!(integration_field_name: field, namely_field_name: to)
+    end
+  end
+
   def integration_key
     integration_field_name.underscore.parameterize("_")
   end
--- a/app/models/net_suite/connection.rb
+++ b/app/models/net_suite/connection.rb
@@ -28,7 +28,7 @@ class NetSuite::Connection < ActiveRecord::Base
 
   def attribute_mapper
     AttributeMapperFactory.new(attribute_mapper: super, connection: self).
-      build_with_defaults { default_field_mappings }
+      build_with_defaults { |mappings| map_defaults(mappings) }
   end
 
   def ready?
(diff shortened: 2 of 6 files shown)


app/models/field_mapping.rb
```diff
@@ -5,6 +5,12 @@ class FieldMapping < ActiveRecord::Base
   validates :attribute_mapper_id, presence: true
   validates :integration_field_name, presence: true
 
+  def self.map!(field, to: nil)
+    unless where(integration_field_name: field).exists?
```
tabfugnic: `unless` seems a bit weird here, thoughts on doing

```ruby
if where(integration_field_name: field).none?
```

app/models/net_suite/connection.rb
```diff
@@ -28,7 +28,7 @@ def attribute_mapper?
 
   def attribute_mapper
     AttributeMapperFactory.new(attribute_mapper: super, connection: self).
-      build_with_defaults { default_field_mappings }
+      build_with_defaults { |mappings| map_default_defaults(mappings) }
```
tabfugnic: `map_default_defaults`
I get what you are trying to do, just took me a minute

Thoughts on just calling it `map_defaults`?

app/models/net_suite/connection.rb
```diff
@@ -66,13 +66,14 @@ def normalizer
     )
   end
 
-  def default_field_mappings
-    remote_fields.merge!(premapped_fields)
+  def map_default_defaults(mappings)
```
tabfugnic: this doesn't seem to be completely related to netsuite and seems a bit more broad. Is this the first sweep or is it possible to factor this out?

spec/models/attribute_mapper_factory_spec.rb
```diff
@@ -29,7 +29,12 @@
           connection: connection
         )
 
-        result = factory.build_with_defaults { { "firstName" => "first_name" } }
+        result = factory.build_with_defaults do |mappings|
+          mappings.create!(
+            integration_field_name: "firstName",
+            namely_field_name: "first_name"
```
tabfugnic: trailing comma?

app/models/field_mapping.rb
jferris: Fixed in 36ab2e8.

app/models/net_suite/connection.rb
jferris: Oops - that was actually a typo. Fixed in a6c4a25.

tabfugnic:
Looks really good. Merge when ready

app/models/net_suite/connection.rb
jferris: Can you explain what you mean? These are the mappings which are supported by NetSuite. Each integration will have different field names.

PR #168, opened 2015-08-18
Store an ID as well as the user-facing field name

jferris:
This is a work in progress.

Because:

* NetSuite uses numeric, account-specific IDs for custom fields
* We want to display friendly labels for mapped fields
* We want to be able to map custom fields

This commit:

* Adds a new column for storing server IDs
* Uses both the friendly column as well as the ID column


kenyonj:
LGTM so far (if it's still a WIP)

iwz:
:+1:

jferris:
I forgot to remove the work in progress line. It's actually complete.

PR #169, opened 2015-08-19
Map custom NetSuite fields

jferris:
Because:

* Users can create custom fields in NetSuite
* Users can export Namely profiles to NetSuite employees

This commit:

* Extends NetSuite scraping to discover custom fields
* Extend NetSuite export to send custom fields
* Allows users to select NetSuite text fields for mapping
* Scrapes beyond the first employee for custom fields

https://trello.com/c/Z2SKrJeU

### DIFF
--- a/app/models/field_mapping.rb
+++ b/app/models/field_mapping.rb
@@ -15,6 +15,14 @@ class FieldMapping < ActiveRecord::Base
     end
   end
 
+  def self.each_with_namely_field
+    all.each_with_object({}) do |field_mapping, result|
+      if field_mapping.namely_field_name?
+        yield field_mapping, result
+      end
+    end
+  end
+
   def integration_key
     integration_field_name.underscore.parameterize("_")
   end
--- a/spec/fixtures/api_responses/net_suite_employee.json
+++ b/spec/fixtures/api_responses/net_suite_employee.json
@@ -43,6 +43,11 @@
                     "internalId": "3992",
                     "scriptId": "custentity_2663_payment_method",
                     "value": false
+                },
+                {
+                    "internalId": "2096",
+                    "scriptId": "custentity_linkedin_url",
+                    "value": ""
                 }
             ]
         },
--- a/spec/models/net_suite/normalizer_spec.rb
+++ b/spec/models/net_suite/normalizer_spec.rb
@@ -2,19 +2,9 @@ require "rails_helper"
 
 describe NetSuite::Normalizer do
   let(:configuration) { double("configuration", subsidiary_id: "123") }
-  let(:normalizer) do
-    employee_json =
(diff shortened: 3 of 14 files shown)


app/models/field_mapping.rb
```diff
@@ -15,6 +15,14 @@ def self.map!(id, name: nil, to: nil)
     end
   end
 
+  def self.map_each
```
drapergeek: What do you think about a name here that implies the check for the `namely_field_name?`? This looks so similar to `each_with_object` that my first guess would be that this is just a ruby/rails method and not dig much further.

spec/models/net_suite/normalizer_spec.rb
```diff
@@ -129,9 +119,63 @@
...
+        export_attributes = export(profile_data, normalizer: normalizer)
+
+        expect(export_attributes["customFieldList"]).to include(
+          "customField" => include(
+            {
+              "internalId" => "1234",
+              "scriptId" => "linkedin_url",
+              "value" => "http://example.com/linkedin"
+            },
+            {
```
jferris: Although the braces aren't necessary here, I'm going to leave them to keep each parameter consistent, as this is actually a variable list of hashes, and formatting just one of them differently would make it appear special.

spec/fixtures/api_responses/net_suite_employee.json
```diff
@@ -43,6 +43,11 @@
                     "internalId": "3992",
                     "scriptId": "custentity_2663_payment_method",
                     "value": false
+                },
+                {
```
drapergeek: What do you think about moving this from a flat file to a method that you can pass in data? How much has the separation of fixtures caused confusion in the specs?

jferris: I'm torn on this one.

I would generally prefer to use factories/generators over fixtures, but fixtures are the way all the specs are implemented. At this point, I'm not sure it would be wise to attempt to replace them all with factories, as it's very difficult to find every relevant part of the fixture, so we'd likely be reducing overall regression coverage.

I also hesitate to use factories for just new content, as I worry that the dual approaches will be more confusing than the fixtures are by themselves.

drapergeek: That's fair.

Proposed option that you can feel free to ignore: moving forward, build each flatfile as a method with default options. That way if you decide to convert them to be factories that can be modified you have the option of doing it without a rewrite but you don't really have to change anything existing?

If you're not feeling pain, it's all good. I just know that the few times I've reviewed the PRs it makes life harder.

drapergeek:
LGTM

app/models/field_mapping.rb
jferris: Renamed in 1e0c1da7.

PR #170, opened 2015-08-19
Always normalize nested fields like job title

jferris:
Because:

* Users can map any Namely field to any NetSuite field
* Fields like job title require converting to send a string
* Mapping job title to another field causes a crash

This commit:

Removes the assumption that job title in Namely is always
mapped to title in NetSuite

https://trello.com/c/NmSrVVJi

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -1,5 +1,4 @@
 class Profile
-  delegate :[], to: :namely_profile
   delegate :update, to: :namely_profile
 
   def initialize(namely_profile)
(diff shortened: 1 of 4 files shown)


app/models/profile.rb
```diff
@@ -10,7 +9,20 @@ def name
...
 
+  def [](key)
+    flatten_hash namely_profile[key]
+  end
+
   private
 
+  def flatten_hash(value)
+    if value.respond_to?(:to_hash)
+      key = value.to_hash.except(:id, "id").keys.first
```
teoljungberg: does this expect the keys to be sorted in a specific way? Can that be conveyed somehow?

jferris: There are reference objects that are returned as an object with `id` and something else, which varies depending on the object. We want the value of the other key. This means that sorting order doesn't really matter, but it does kind of assume that there are only two keys.

I agree that the current implementation isn't very clear, but I'm sort of at a loss as to how to clean it up. Any suggestions?

teoljungberg: What different variation of keys for the name could we have? How about having a whitelist for those and getting the one who appear first?

jferris: I tried a simpler approach in 376ebc2. Thoughts?

teoljungberg: I like how 376ebc2 looks

teoljungberg:
LGTM, if the build goes green

PR #171, opened 2015-08-21
Remove rack-timeout from development, test

jferris:
Because:

* Integration tests are randomly hanging
* rack-test interacts strangely with threads

This commit:

* Removes rack-timeout from development and test

https://trello.com/c/sMfXD1IQ


derekprior:
:ship:

PR #172, opened 2015-08-21
Fix randomly failing spec

jferris:
These specs were failing because time could proceed one second between
when the digest was calculated and when it was checked.


derekprior:
:+1:

PR #173, opened 2015-08-24
Add Cloud Elements keys to .sample.env

jferris:
These keys are required to pass tests and boot a development
environment.

Actual values are required to connect to NetSuite.


derekprior:
:+1:

PR #174, opened 2015-08-24
Set environment variables before booting the app

derekprior:
If the HOST environment variable is not set, `docker-compose up` will
not completely successfully.


jferris:
Looks good!

PR #175, opened 2015-08-24
Improve NetSuite field mapping list

jferris:
* Blacklist unmappable NetSuite fields
* Display mappable fields alphabetically
* Premap additional NetSuite fields
* Sort Namely fields alphabetically

https://trello.com/c/sxr8CVBq

### DIFF
--- a/app/models/field_mapping.rb
+++ b/app/models/field_mapping.rb
@@ -23,6 +23,10 @@ class FieldMapping < ActiveRecord::Base
     end
   end
 
+  def self.ordered
+    order(integration_field_name: :asc)
+  end
+
   def integration_key
     integration_field_name.underscore.parameterize("_")
   end
--- a/app/models/net_suite/connection.rb
+++ b/app/models/net_suite/connection.rb
@@ -88,6 +88,9 @@ class NetSuite::Connection < ActiveRecord::Base
     mappings.map! "firstName", to: "first_name", name: "First name"
     mappings.map! "gender", to: "gender", name: "Gender"
     mappings.map! "lastName", to: "last_name", name: "Last name"
+    mappings.map! "middleName", to: "middle_name", name: "Middle name"
+    mappings.map! "mobilePhone", to: "mobile_phone", name: "Mobile phone"
+    mappings.map! "officePhone", to: "office_phone", name: "Office phone"
     mappings.map! "phone", to: "home_phone", name: "Phone"
     mappings.map! "title", to: "job_title", name: "Title"
   end
--- a/app/models/net_suite/employee_fields_loader.rb
+++ b/app/models/net_suite/employee_fields_loader.rb
@@ -13,14 +13,19 @@ module NetSuite
 
       employees.
         flat_map { |employee| profile_fields_from_employee_hash(employee) }.
-        uniq(&:id)
+        uniq(&:id).
+        reject { |field| blacklist.include?(field.id) }
+    end
+
+    private
+
+    def blacklist
+      ENV.fetch("NET_SUITE_FIELD_BLACKLIST", "").split(",")
     end
 
     def profile_fields_from_employee_hash(employee)
       standard_fields(employee) + custom_fields(employee)
     end
-
-    private
 
     def standard_fields(employee)
       employee.except("customFieldList").map do |id, value|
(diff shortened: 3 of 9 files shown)


app/models/net_suite/employee_fields_loader.rb
```diff
@@ -13,15 +13,20 @@ def load_profile_fields
...
         flat_map { |employee| profile_fields_from_employee_hash(employee) }.
-        uniq(&:id)
+        uniq(&:id).
+        reject { |field| blacklist.include?(field.id) }
+    end
+
+    private
+
+    def blacklist
+      ENV.fetch("NET_SUITE_FIELD_BLACKLIST", "").split(",")
```
tabfugnic: Any reason why this black list is an environment variable?

halogenandtoast:
Looks good to me.

app/models/net_suite/employee_fields_loader.rb
jferris: We expect this to change frequently, and it would be nice to avoid a deploy for each new field.

app/models/user.rb
```diff
@@ -17,6 +17,7 @@ def namely_fields_by_label
     namely_fields.
       all.
       select { |field| AttributeMapper::SUPPORTED_TYPES.include?(field.type) }.
+      sort_by { |field| field.label.downcase }.
```
tabfugnic: Thoughts on using `order` before `all` instead?

jferris: This is a `Namely::Collection` from the Namely Ruby client, and not an ActiveRecord relation; it doesn't respond to `order`.

app/models/field_mapping.rb
```diff
@@ -23,6 +23,10 @@ def self.each_with_namely_field
     end
   end
 
+  def self.ordered
+    order(:integration_field_name)
```
JoelQ: Thoughts on explicitly setting `:asc` or `:desc` here?

tabfugnic:
Looks good

app/models/net_suite/connection.rb
```diff
@@ -88,6 +88,9 @@ def map_standard_fields(mappings)
     mappings.map! "firstName", to: "first_name", name: "First name"
     mappings.map! "gender", to: "gender", name: "Gender"
     mappings.map! "lastName", to: "last_name", name: "Last name"
+    mappings.map! "middleName", to: "middle_name", name: "Middle name"
```
teoljungberg: Thoughts on having these :abc: sorted?

jferris: I believe they're alphabetical already?

app/models/user.rb
```diff
@@ -17,6 +17,7 @@ def namely_fields_by_label
     namely_fields.
       all.
       select { |field| AttributeMapper::SUPPORTED_TYPES.include?(field.type) }.
```
JoelQ: Is this `Enumerable#select` or `ActiveRecord#select` ? If it is `Enumerable#select`, thoughts on using `ActiveRecord#where` instead?

app/models/net_suite/connection.rb
teoljungberg: They are, my bad

spec/models/field_mapping_spec.rb
```diff
@@ -93,6 +93,21 @@
...
+    it "returns field mappings in alphabetical order" do
+      map_fields(
+        "fieldB" => "b",
+        "fieldA" => "a",
+        "fieldC" => "c",
+      ).each(&:save!)
+
+      result = FieldMapping.ordered
+
+      expect(result.map(&:integration_field_name)).
```
teoljungberg: thoughts on pulling out `result.map(....)` into a variable for clarity?

app/models/field_mapping.rb
jferris: Done in 5eb1b20.

app/models/user.rb
jferris: It's a `Namely::Collection`, from the Ruby API client. It doesn't respond to `where`.

spec/models/field_mapping_spec.rb
jferris: Would you name the variable `integration_field_names`? My feeling is that the extraction would add indirection, but not clarity.

PR #176, opened 2015-08-25
Generate better instance names on Cloud Elements

derekprior:
Previously, we had many conflicting instance names because they were
keyed only on the account id. This was confusing in development, but
potentially dangerous as we rolled out to other environments where the
instance names would conflict yet again.

This change adds both the Rails environment and a timestamp to the name
to help make finding the correct cloud elements instance easier.

### DIFF
--- a/app/models/net_suite/authentication.rb
+++ b/app/models/net_suite/authentication.rb
@@ -25,7 +25,7 @@ module NetSuite
     private
 
     def create_instance
-      result = @client.create_instance(attributes)
+      result = @client.create_instance(self)
       if result.success?
         @connection.update!(
           instance_id: result[:id],
--- a/app/models/net_suite/client.rb
+++ b/app/models/net_suite/client.rb
@@ -33,21 +33,11 @@ module NetSuite
       )
     end
 
-    def create_instance(params)
+    def create_instance(authentication)
       submit_json(
         :post,
         INSTANCES,
-        "configuration" => {
-          "user.username" => params[:email],
-          "user.password" => params[:password],
-          "netsuite.accountId" => params[:account_id],
-          "netsuite.sandbox" => false
-        },
-        "element" => {
-          "key" => "netsuiteerp"
-        },
-        "tags" => [],
-        "name" => "#{params[:account_id]}_netsuite"
+        Instance.new(authentication).to_h
       )
     end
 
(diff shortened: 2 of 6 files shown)


spec/models/net_suite/instance_name_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+module NetSuite
```
jferris: Bold! I haven't taken this approach for a few years.

I'm not opposed to it in general, but it's not the style in use by most of the other specs. Do you think it's worth the inconsistency?

app/models/net_suite/instance_name.rb
```diff
@@ -0,0 +1,7 @@
+module NetSuite
+  class InstanceName
+    def self.new(account_id)
```
jferris: This strikes me as confusing. It's unusual to define `new`, and the implementation means that `InstanceName.new` won't return an instance of `InstanceName`. What do you think about defining `to_s` to return the name instead?

jferris: Alternatively, you could pass in a `Connection` instance and move the whole `to_json` block for instance creation onto this class, making a `NetSuite::Instance` class.

spec/models/net_suite/instance_name_spec.rb
derekprior: Oh, this is just a habbit I got into because I once got tired of namespacing every reference to the described class. I can switch it. I don't care strongly.

app/models/net_suite/instance_name.rb
derekprior: I like the second suggestion.

I'm okay with `new` not returning an instance of the thing you are `new`ing in simple cases. It's a `String` factory. But with the second approach, this is unnnecessary.

derekprior:
@jferris Ready for another look. Passing the `authentication` and extracting the Instance object ended up highlighting an actual collaborator in the client specs. We were sweeping that under the rug by being able easily pass in `{}` where we didn't care. I don't love ancilary client spec changes and am open to suggestions there, certainly.

jferris:
This looks good to me.

> I don't love ancilary client spec changes and am open to suggestions there, certainly.

I'd chalk that up to an Extract Parameter Object refactoring.

PR #177, opened 2015-08-25
Add task for clearing local field mappings

jferris:
We need to do this a lot when debugging locally so that we can test out uncached UI interactions.


derekprior:
Ship away...

PR #179, opened 2015-08-25
Only fetch Namely fields once

jferris:
We were fetching the list of Namely fields from the API for each
mappable field.

This commit changes the view to only perform that request once.

Locally, this increased performance for this page by about 30x.


derekprior:
:fast_forward:

PR #180, opened 2015-08-25
Discard partial fields if NetSuite fails

jferris:
Because:

* We create standard fields before fetching custom fields from NetSuite
* NetSuite's API sometimes fails unexpectedly
* We only look for custom fields if no fields have been mapped

This commit:

* Creates the attribute mapper in a database transaction
* Performs field mapping in the same transaction

So that:

* A failed fetch will rollback the transaction
* Subsequent requests will try again from scratch

https://trello.com/c/Eykx3oI7


derekprior:
:+1:

PR #181, opened 2015-08-26
Allow editing of configuration and field mappings

derekprior:
Previous to this change, users with existing connections could only
edit authentication information. They were unable to edit existing
mappings or to change any additional configuration information
(i.e. NetSuite subsidiary).

This change adds distinct edit links for configuration and mapping
where applicable.

### DIFF
--- a/spec/features/user_edits_existing_field_mappings_spec.rb
+++ b/spec/features/user_edits_existing_field_mappings_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+feature "User edits existing field mappings" do
+  scenario "successfully" do
+    personal_email = { text: "Personal email", value: "personal_email" }
+    stub_namely_fields("fields_with_jobvite")
+    user = create(:user)
+    create(
+      :jobvite_connection,
+      :connected,
+      :with_namely_field,
+      installation: user.installation
+    )
+
+    visit dashboard_path(as: user)
+    click_jobvite_mappings_link
+    first_mappable_field.select(personal_email[:text])
+    click_button t("attribute_mappings.edit.save")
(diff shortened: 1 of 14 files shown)


spec/support/features/net_suite_requests_helper.rb
```diff
@@ -0,0 +1,18 @@
+module Features
+  def stub_lookup_subsidiaries(status:, body:)
```
jferris: Should this method name have `net_suite` in there somewhere?

spec/features/user_edits_existing_field_mappings_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def click_jobvite_mappings_link
+    within(".jobvite-account") do
+      click_link t("dashboards.show.edit_mappings")
+    end
+  end
+
+  def first_mappable_field
+    page.all("select").first
```
jferris: This will immediately return `nil` if it hasn't finished rendering yet in an asynchronous test, which would mean this test becomes erratic once we flip JavaScript on.

You can use `find("select", match: :first)` instead, which will wait until a `select` element appears and then return it (or raise an exception after a timeout).

https://robots.thoughtbot.com/write-reliable-asynchronous-integration-tests-with-capybara

jferris:
I had a couple minor comments. Ready to merge whenever you're feeling good about it.

spec/features/user_edits_existing_field_mappings_spec.rb
derekprior: Ah. I thought they were equivalent and I can never remember the option name for the `:first` value. Thanks.

PR #182, opened 2015-08-26
Fix exception in NetSuite background sync

jferris:
We changed the internal API for this class, but there wasn't an
integration test to ensure API compatibility.

I extracted the particular piece that changed to a class method and
added a test to prevent a similar regression.

https://trello.com/c/k0OIdkkF


derekprior:
:+1:

PR #183, opened 2015-08-26
Users can select mappings for Greenhouse accounts

jferris:
Because:

* Greenhouse contains fields that Namely doesn't
* Users can create custom fields in Namely for those fields

This commit:

* Adds attribute mapping for Greenhouse connections
* Maintains hard mappings for address and salary, which we can't map yet

https://trello.com/c/kYjN3NrG

### DIFF
--- a/app/models/greenhouse/normalizer.rb
+++ b/app/models/greenhouse/normalizer.rb
@@ -1,116 +1,136 @@
 module Greenhouse
   class Normalizer
-    # Initializes an attribute mapper using Namely Fields
-    def initialize(namely_fields)
+    NAMELY_IDENTIFIER_FIELD = :greenhouse_id
+
+    def initialize(attribute_mapper:, namely_fields:)
+      @attribute_mapper = attribute_mapper
       @namely_fields = namely_fields
     end
 
-    def call(payload)
-      custom_fields(candidate(payload)).
-        merge(custom_fields(offer(payload))).
-        merge(custom_fields(job(payload))).
-        merge(basic_fields(payload)).
-        select { |_, value| value.present? }
+    def call(data)
+      candidate = Payload.new(data, namely_fields: @namely_fields).to_hash
+      imported = @attribute_mapper.import(candidate)
+      imported.merge(
+        candidate.slice(NAMELY_IDENTIFIER_FIELD, :salary, :home, :user_status)
+      )
     end
 
     def namely_identifier_field
-      :greenhouse_id
+      NAMELY_IDENTIFIER_FIELD
     end
 
     private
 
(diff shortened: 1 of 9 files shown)


app/models/greenhouse/normalizer.rb
```diff
@@ -1,116 +1,137 @@
...
 
-    def call(payload)
-      custom_fields(candidate(payload)).
-        merge(custom_fields(offer(payload))).
-        merge(custom_fields(job(payload))).
-        merge(basic_fields(payload)).
-        select { |_, value| value.present? }
+    def call(data)
+      candidate = Payload.new(data, namely_fields: @namely_fields).to_hash
+      exported = @attribute_mapper.import(candidate)
```
derekprior: I'm a bit confused as to why the result of `import`ing something would be called `exported`?

jferris: I was originally calling `export` here until I remembered this is supposed to be an import. I forgot to rename the variable, though.

Fixed in 8b32d2f.

jferris:
I'm going to leave the redundant empty else clauses, because that's not part of the styleguide we're using, the explicit `nil` returns add clarity in this case, and this is existing code I re-indented.

app/models/greenhouse/normalizer.rb
```diff
@@ -1,116 +1,136 @@
...
-      home_address = candidate.fetch("addresses", [])
-      if home_address
-        home_address = home_address.find do |address|
-          address.fetch("type") == "home"
-        end || {}
-        { address1: home_address.fetch("value", "") }
-      else
-        nil
+      def home_address
+        home_address = candidate.fetch("addresses", [])
```
derekprior: What do you think about `candidate.fetch("addresses", nil)` (or just direct has access) instead? As currently written, I think your if statement never evaluates to false... unless the value of "addresses" is falsey?

Then I think you could get rid of the `else` branch too?

derekprior:
looks good.

app/models/greenhouse/normalizer.rb
jferris: It's actually different. This actually returns `nil`:

``` ruby
{ "addresses" => nil }.fetch("addresses", [])
```

That means this method can return any of the following:

```
nil
{ address1: "" }
{ address1: nil }
{ address1: "Address" }
```

This logic does seem overly defensive to me, but I haven't found a different combination which passes the tests. I'll take a look at the API to see if the tests are too paranoid.

jferris: These appear to all be possible edge cases in the real world, so I'm going to leave the existing logic.

PR #184, opened 2015-08-26
Map namely user status to NetSuite inactive flag

derekprior:
This is a field that users are likely to want synchronized between the
two systems, so we've added a default mapping for it. The mapping must
be normalized because the types differ between the systems.


app/models/net_suite/normalizer.rb
```diff
@@ -29,6 +32,10 @@ def map_gender(value)
     GENDER_MAP[value]
   end
 
+  def map_user_status(value)
+    value == "inactive".freeze
```
jferris: Since this string is immediately constructed and discarded each time, is there a benefit to freezing it?

derekprior: not sure how that got in there.

jferris:
Besides that question about the frozen string, this looks good to me.

PR #185, opened 2015-08-27
Fix path to header logo

derekprior:
The `image-url` helper is asset pipeline aware. Prior to this change,
this image was causing an `ActionController::RoutingError (No route
matches [GET] "/assets/logo_full.svg")` error.

### DIFF
--- a/app/assets/stylesheets/_layout.scss
+++ b/app/assets/stylesheets/_layout.scss
@@ -19,7 +19,7 @@ header {
     padding-top: 5px;
 
     .icon {
-      background: transparent url('logo_full.svg') no-repeat;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/_layout.scss
```diff
@@ -19,7 +19,7 @@ header {
     padding-top: 5px;
 
     .icon {
-      background: transparent url('logo_full.svg') no-repeat;
+      background: transparent image-url('logo_full.svg') no-repeat;
```
jferris: I don't think the first part of this is good advice.

jferris:
This looks good to me.

PR #186, opened 2015-08-27
Allow NetSuite to operate without a subsidiary

jferris:
Because:

* Users can setup NetSuite without subsidiaries

This commit:

* Checks to see if NetSuite is configured with subsidiaries
* Skips subsidiary configuration when none are present

Side effects:

* Adds a cache column to prevent repeated NetSuite requests

https://trello.com/c/43U5qp9V


derekprior:
`NetSuite::Connection.configurable?` should be updated to delegate to this flag as well.

derekprior:
Looks good.

PR #187, opened 2015-08-27
Configure Raygun for exception tracking

derekprior:
There are some exceptions occurring on staging and we're not noticing
them because we're not always looking at the logs.


jferris:
This looks good to me.

PR #189, opened 2015-08-27
Allow log level to be changed via ENV

derekprior:
This will allow us to more quickly get debugging data back from deployed
environments.


jferris:
Looks good to me.

PR #190, opened 2015-08-31
Map Namely departure date to NetSuite release date

derekprior:
These are the names for the termination date concept in each system.
CloudElements/NetSuite require milliseconds since epoch, so we need to
normalize for that.

Issues to follow up on:

* Can't clear release date in NetSuite once it is set. Neither setting
  to empty string nor nil works. It's unclear if this is a NetSuite or
  CloudElements issue.
* Setting departure date as, for example, 09/05/2015 in Namely causes it
  to sync to NetSuite as milliseconds since the epoch (UTC) at midnight
  of that day. When displayed in NetSuite it seems some time zone
  conversion is done, causing the date to be displayed as 09/04/2015.
  Not sure if there's a sensible thing to do here.

### DIFF
--- a/app/models/fields/date_value.rb
+++ b/app/models/fields/date_value.rb
@@ -16,7 +16,11 @@ module Fields
     end
 
     def to_date
-      DateTime.strptime(@value, DATE_FORMAT).to_date
+      if @value.present?
+        DateTime.strptime(@value, DATE_FORMAT).to_date
+      end
+    rescue ArgumentError
+      nil
     end
 
     def to_address
--- a/app/models/fields/null_value.rb
+++ b/app/models/fields/null_value.rb
@@ -1,0 +1,19 @@
+module Fields
+  class NullValue
+    def to_raw
+      nil
+    end
+
+    def to_s
+      nil
+    end
+
+    def to_date
+      nil
+    end
+
+    def to_address
+      nil
+    end
+  end
+end
--- a/app/models/net_suite/normalizer.rb
+++ b/app/models/net_suite/normalizer.rb
@@ -24,10 +24,12 @@ class NetSuite::Normalizer
     end
 
     def to_hash
-      mapped_attributes.
-        merge(address_attributes).
-        merge(string_attributes).
-        merge(custom_fields_attributes)
+      with_null_field_list(
+        mapped_attributes.
+          merge(address_attributes).
+          merge(string_attributes).
+          merge(custom_fields_attributes)
+      )
     end
 
     private
@@ -147,6 +149,14 @@ class NetSuite::Normalizer
       if @attributes["address"]
         @attributes["address"].to_address
       end
+    end
+
+    def with_null_field_list(fields)
+      fields.tap do |mapped_fields|
+        mapped_fields["nullFieldList"] = mapped_fields.select do |_, value|
+          value.nil?
+        end.keys
+      end.compact
     end
   end
 
--- a/app/models/net_suite/normalizer.rb
+++ b/app/models/net_suite/normalizer.rb
@@ -37,6 +37,7 @@ class NetSuite::Normalizer
         "gender" => gender,
         "isInactive" => user_status,
         "subsidiary" => subsidiary,
+        "releaseDate" => release_date,
       }
     end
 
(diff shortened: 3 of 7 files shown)


jferris:
The code looks good to me.

For the UTC issue, does it make sense for us to convert the date to UTC before sending? Do we know which timezone the input date is in?

derekprior:
The date is in UTC when we send it. When I pull it out of the API, it's 5 hours earlier (aligns with current US Eastern Time). I checked the preferences for our NetSuite user and it was set to -8 hours. Just for kicks I set it to GMT, re-ran the sync, and still the same result.

I assume this is a NetSuite setting. I don't have any idea how we would know what time zone to send date/times in unless we asked for it as part of connection setup.

app/models/fields/date_value.rb
```diff
@@ -16,7 +16,17 @@ def to_s
...
-      DateTime.strptime(@value, DATE_FORMAT).to_date
+      if @value.present?
+        DateTime.strptime(@value, DATE_FORMAT).to_date
+      end
+    rescue ArgumentError
+      nil
+    end
+
+    def to_i
+      if to_date.present?
```
jferris: Do we actually want this conversion, or would `nil` make more sense?

app/models/net_suite/normalizer.rb
```diff
@@ -24,10 +24,12 @@ def initialize(attributes, subsidiary_id)
     end
 
     def to_hash
-      mapped_attributes.
-        merge(address_attributes).
-        merge(string_attributes).
-        merge(custom_fields_attributes)
+      with_null_field_list do
```
jferris: Thoughts on following the same pattern as the others with `merge(null_attributes)`? I don't think one is really better than the other, but the block form makes it seem more different to me than it is.

derekprior: That was my original thought as well, but I need access to the currently-being-built hash to detect what is nil. It feels like I should do that *after* any other normalization has been done. It could be a method:

```
with_null_field_list(
  mapped_attributes.
    merge(...).
    merge(...).
    merge(...)
)
```
But at that point the block form felt like it was more descriptive to me. But I don't feel strongly about that.

app/models/fields/date_value.rb
derekprior: This is used in setting the date to ms from epoch in the normalizer. `to_i` mapping to seconds from epoch felt more consistent with ruby's `to_i` method on DateTime. And the MS bit felt like ti made more sense in the normalizer itself.

app/models/net_suite/normalizer.rb
jferris: Oh, I see; yes, I think an argument makes more sense than a block in that case.

app/models/fields/date_value.rb
jferris: Okay - I could go either way on this.

Using `to_i` directly in the NetSuite normalizer will make it create strange dates in NetSuite for invalid mappings as long as the type can be converted to an integer. That could be surprising with string fields, since `String#to_i` returns `0`.

app/models/net_suite/normalizer.rb
```diff
@@ -65,6 +68,14 @@ def user_status
       @attributes["isInactive"].to_s == "inactive"
     end
 
+    def release_date
+      date = @attributes.fetch("releaseDate", Fields::NullValue.new).to_date
```
derekprior: this could also be another level of nesting:

```ruby
if @attributes["releaseDate"]
  date = @attributes["releaseDate"].to_date

  if date.present?
    # calculation
  end
end
```
thoughts?

app/models/fields/null_value.rb
```diff
@@ -0,0 +1,19 @@
+module Fields
+  class NullValue
```
jferris: Nice. I was thinking of doing this earlier.

I know there are some explicit checks for `if field`, `if field.present?`, and similar. Might be worth a quick audit.

jferris: Oh, looks like it's only used in one place - nevermind.

app/models/net_suite/normalizer.rb
jferris: I like the approach with `NullValue`. It could be a nice refactoring to make the attribute mapper return those instead of `nil` for nulls in the source data. I think that makes sense as a separate pull request, though.

jferris:
This looks good to me.

PR #191, opened 2015-08-31
Improve dashboard user experience

derekprior:
Clean up dashboard display

* Move edit actions into a readable list
* Make all button styles consistent
* Move disconnect button into header to mirror "connect"

Add "Show Webhook URL" buttons where appropriate

Having a random URL hanging around the UI for iCIMS and Greenhouse doesn't
make a lot of sense and is also not useful to the user most of the time. By
tucking it behind a button, we can both better label it and keep it from
cluttering the UI in the majority of cases where no one needs to see it.

This is added progressively so that users without JavaScript enabled will
see the labeled webhook at all times, without a click. They will not see
the button.

https://trello.com/c/vKySDBNc


derekprior:
dashboard

app/views/dashboards/_connected_connection.html.erb
```diff
@@ -1,28 +1,39 @@
-<h3>
-  <%= fa_icon(
-    "check-circle",
-    class: "connected",
-    text: t("#{connection.integration_id}.name")
-  ) %>
-</h3>
+<div class="connection-header">
```
jferris: Is there something more semantic we can use here like `section` and `header`?

jferris:
This looks like a big improvement to me. I think there may be some more semantic HTML elements available if you think it's worth playing with them. I think this is good to merge whenever you're happy with the markup.

app/views/dashboards/_connected_connection.html.erb
derekprior: `header` would work here, but it requires overriding some other CSS we're inheriting from somewhere else. In the interest of moving this along, I'm going to roll with this for now.

PR #192, opened 2015-08-31
Refactor type conversion for profile values

jferris:
Because:

* We need to export complex types like addresses and references
* The format for those types varies between services

This commit:

* Extends our type conversion to track Namely types

Notes:

* This refactoring is a prerequisite for properly exporting addresses
* This doesn't modify imports, as those are not relevant for NetSuite

https://trello.com/c/7ucossFw

### DIFF
--- a/app/models/fields/collection.rb
+++ b/app/models/fields/collection.rb
@@ -1,0 +1,40 @@
+module Fields
+  # Wraps values from Namely profile fields in an object able to convert to
+  # useful formats.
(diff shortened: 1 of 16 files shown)


derekprior:
I like this. I think it would be best if we added a few lines of rdoc to the top of each of the fields classes (and eventually the other classes). I think that could go a long way to easing the documentation concern from alst weeks retro too.

PR #193, opened 2015-09-01
Export address fields to NetSuite

jferris:
Because:

* Namely and NetSuite support address fields

This commit:

* Adds support for exporting address fields from Namely
* Adds a default field mapping for the home address

https://trello.com/c/7ucossFw


app/models/net_suite/constants.rb
```diff
@@ -0,0 +1,259 @@
+module NetSuite
+  module Constants
```
jferris: I'm not in love with this, but it was pretty distracting to have it at the top of the class that used these values.

I could move them into JSON or something, but that would just be harder to follow.

derekprior:
:+1:

PR #194, opened 2015-09-01
Update Rails and RSpec

derekprior:
Rails is several security versions out of date. I originally
updated to 4.1.13, which worked fine. It proved to be very little
work to go ahead and jump to 4.2.4. With Rails 5 coming this year,
it seems prudent to stay current.

I updated RSpec as well to give us access to potentially useful
new features in 3.3, such as `rspec bisect`.


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: git://github.com/circleci/rspec_junit_formatter.git
-  revision: 7b098c8ab2ac5e5f78190b51629ed6d67a9de4d1
+  revision: 6885218137afbadf560c922114767866953f8cbf
```
jferris: While we're updating this: are we using the "junit" formatter for RSpec?

jferris:
This looks good to me.

Gemfile.lock
derekprior: @macat added it back in December. It does produce an XML artifact on Circle. Not sure if that's used somehow? @macat?

PR #195, opened 2015-09-01
Use absolute reference for constant

jferris:
Although this reference worked in development and test, production-ready
eager loading wasn't able to load the correct file.


derekprior:
:open_mouth:

PR #196, opened 2015-09-02
Add activity feed for NetSuite syncs

derekprior:
This gives users a web-accessible view of the sync activity happening
with their net suite integration. The data is similar to what is emailed
to them.

The view currently does not distinguish between successful or
unsuccessful profile events, but we have a follow-on task that will take
care of that.

https://trello.com/c/WsHL9kNH

### DIFF
--- a/app/models/sync_notifier.rb
+++ b/app/models/sync_notifier.rb
@@ -1,0 +1,49 @@
+class SyncNotifier
+  def self.deliver(installation:, integration_id:, results:)
+    new(
+      installation: installation,
+      integration_id: integration_id,
+      results: results
+    ).deliver
+  end
+
+  def initialize(installation:, integration_id:, results:)
+    @installation = installation
+    @integration_id = integration_id
+    @results = results
+  end
+
+  def deliver
+    deliver_emails
+    record_sync_summary
+  end
+
+  private
+
+  attr_reader :installation, :integration_id, :results
+
+  def deliver_emails
+    users.each do |user|
+      SyncMailer.sync_notification(
+        email: user.email,
+        integration_id: integration_id,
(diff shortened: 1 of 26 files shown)


app/models/sync_notifier.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def initialize(installation:, integration_id:, results:)
+    @installation = installation
+    @integration_id = integration_id
+    @results = results
+  end
+
+  def deliver
+    deliver_emails
+    record_activity
```
derekprior: record_sync_summary

app/controllers/activities_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ActivitiesController < IntegrationController
+  def index
```
jferris: Should this be `SyncsController#index`?

db/schema.rb
```diff
@@ -11,126 +11,145 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150827155439) do
+ActiveRecord::Schema.define(version: 20150902134111) do
```
jferris: Do we want to update this on another branch to filter out the Rails upgrade changes?

jferris: Pull request #197 addresses this.

app/controllers/activities_controller.rb
derekprior: If we wanted to maintain the symetry of controller and model names then it would be `SyncSummariesController`, right?

We want the user facing consept to still be "activities" or "activity feed", so it boils down to whether we want the mismatch to be in the route or the controller?

PR #197, opened 2015-09-02
Update db/schema.rb for Rails 4.2

derekprior:
Rails 4.2 schema dumper has some changes from Rails 4.1. This change
contains is the result of `rake db:schema:dump` under Rails 4.2.


jferris:
Looks good.

PR #198, opened 2015-09-02
Instantiate SyncJob with a connection

derekprior:
This previously took two arguments, both of which are accessible via the
connection. We had convenient access to the connection in all places
where we were instantiating Sync Jobs.

### DIFF
--- a/app/controllers/syncs_controller.rb
+++ b/app/controllers/syncs_controller.rb
@@ -1,8 +1,5 @@
 class SyncsController < IntegrationController
   def create
-    Delayed::Job.enqueue(
-      SyncJob.new(integration_id, current_user.installation_id)
-    )
-    connection
+    SyncJob.perform_later(connection)
   end
 end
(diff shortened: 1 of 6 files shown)


jferris:
My intention here was to serialize reliable values (IDs, enumerations) instead of trying to serialize ActiveRecord instances (`Connection` instances).

Do you think it's safe to serialize `Connection` instances to YAML? I've run into trouble when trying to do that in the past.

derekprior:
Hm. Maybe. I vaguely remember this. In any event, in the brave new world of 4.2, we can use active job, which will rely on [globalid](https://github.com/rails/globalid) which is meant to avoid this very problem, I think?

We'd just switch up our queuing a tiny bit, I think. Thoughts?

jferris:
I'm open to using active job or globalid. I haven't used that to get around the "YAML mostly doesn't work" issue yet. Do we need to make further changes, or is this already being handled because we're on 4.2?

derekprior:
Looks like relatively minor changes to make this job an "ActiveJob" and enqueue it as such.

jferris:
Okay, let's see this bold new future!

derekprior:
@jferris this is ready for another look.

jferris:
This looks much better!

PR #199, opened 2015-09-02
Fail faster when interacting with Cloud Elements

jferris:
Because:

* The Cloud Elements API can fail during any request
* We don't have explicit handling for many of these failures

This commit:

* Raises exceptions whenever a request fails
* Wraps known exceptions in a special exception class
* Rescues exceptions in the two places we know how to handle them

### DIFF
--- a/app/models/net_suite/export.rb
+++ b/app/models/net_suite/export.rb
@@ -48,18 +48,17 @@ module NetSuite
       end
 
       def update
-        response = @net_suite.update_employee(id.to_s, attributes)
-        Result.new(response, true, @profile)
+        request(updated: true) do
+          @net_suite.update_employee(id.to_s, attributes)
+        end
       end
 
       def create
-        response = @net_suite.create_employee(attributes)
-
-        if response.success?
+        request(updated: false) do
+          response = @net_suite.create_employee(attributes)
           @profile.update(netsuite_id: response["internalId"])
+          response
         end
-
-        Result.new(response, false, @profile)
       end
 
       def id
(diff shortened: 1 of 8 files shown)


app/models/net_suite/export.rb
```diff
@@ -69,29 +68,37 @@ def id
       def attributes
         @normalizer.export(@profile)
       end
+
+      def request(updated:)
+        yield
+        Result.new(true, nil, updated, @profile)
```
derekprior: Seems like a prime candidate for keyword arguments in a future refactor...

jferris: Yeah, I could do that while I'm moving all the arguments around.

derekprior:
:+1:

app/models/net_suite/export.rb
```diff
@@ -69,29 +68,37 @@ def id
...
-        @response = response
+      def initialize(success, message, updated, profile)
+        @success = success
+        @message = message
         @updated = updated
         @profile = profile
       end
 
-      delegate :success?, to: :response
+      attr_reader :message
```
jferris: @derekprior do you think `error` would be a better name for this?

app/models/net_suite/export.rb
jferris: Updated in 965749c.

app/models/net_suite/export.rb
derekprior: Yes. Makes it clear we only expect it to be set in that case.

app/models/net_suite/export.rb
```diff
@@ -69,29 +68,47 @@ def id
...
+          updated: updated,
+          profile: @profile
+        )
+      end
     end
 
     class Result
-      def initialize(response, updated, profile)
-        @response = response
+      def initialize(success:, message:, updated:, profile:)
```
derekprior: should `message` be optional?

jferris: `Result` is only initialized internally from this class, which success and failure cases. It seemed better to explicitly provide `nil` to make it clear that there isn't a `message` in the success case.

(I could also make two different classes here, one of which accepts `message`, and the other which doesn't)

app/models/net_suite/export.rb
jferris: Renamed in bbd1c45.

jferris:
@derekprior ready for re-review.

derekprior:
looks good

PR #200, opened 2015-09-03
Display failure messages for failed profiles

jferris:
When a sync fails to export one or more profiles, record the failure
message from the remote service on the profile event and display it in
the feed.

Also separates out failed profile events into their own section in the
feed.

https://trello.com/c/781vB8tU

### DIFF
--- a/app/models/profile_event.rb
+++ b/app/models/profile_event.rb
@@ -4,10 +4,22 @@ class ProfileEvent < ActiveRecord::Base
   # Intended to be called via the association from a SyncSummary which will
   # provide the required `sync_summary_id` attribute.
   def self.create_from_result!(result)
-    create!(profile_name: result.name, successful: result.success?)
+    create!(profile_name: result.name, error: result.error)
   end
 
   def self.ordered
     order(profile_name: :asc)
+  end
+
+  def self.successful
+    where(error: nil)
+  end
+
+  def self.failed
+    where.not(error: nil)
+  end
+
+  def successful?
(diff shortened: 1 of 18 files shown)


jferris:
Screenshot

app/models/profile_event.rb
```diff
@@ -4,10 +4,22 @@ class ProfileEvent < ActiveRecord::Base
...
   def self.ordered
     order(profile_name: :asc)
   end
+
+  def self.successful
+    where(error: nil)
+  end
+
+  def self.failed
+    where.not(error: nil)
```
derekprior: would `where.not(successful)` work?

db/migrate/20150902145916_add_error_to_profile_events.rb
```diff
@@ -0,0 +1,18 @@
+class AddErrorToProfileEvents < ActiveRecord::Migration
+  def up
+    add_column :profile_events, :error, :string
```
derekprior: Should we index this field? Maybe even conditionally (`where error is null`)?

jferris: Done in cffef58.

app/models/profile_event.rb
jferris: That results in:

    TypeError: Cannot visit ProfileEvent::ActiveRecord_Relation

derekprior:
:+1:

PR #201, opened 2015-09-03
Rename and Refactor AuthenticationNotifier

derekprior:
* It handles `Unauthorized` errors, not authentication errors.
* Like SyncJob, it needs only a connection to operate on.


jferris:
This looks good to me.

PR #202, opened 2015-09-03
Make Unauthorized and Sync notifiers consistent

derekprior:
Make the two notifiers use the same interface:

* class method `deliver`
* instance method `deliver`


jferris:
Nice. Looks good to merge.

PR #203, opened 2015-09-03
Simplify sync notifications

jferris:
Because:

* Failures were preventing the activity feed from updating
* We had duplicate logic for summarization sync results
* Failed emails will result in retrying the entire job

This commit:

* Records sync events before attempting to send emails
* Reuses the existing sync summary for email content
* Sends emails in separate background jobs

### DIFF
--- a/app/mailers/sync_mailer.rb
+++ b/app/mailers/sync_mailer.rb
@@ -1,63 +1,15 @@
 class SyncMailer < ApplicationMailer
-  def sync_notification(email:, integration_id:, results:)
-    @integration_id = integration_id
-    @sync_results = SyncResults.new(integration_id, results)
+  def sync_notification(email:, sync_summary:)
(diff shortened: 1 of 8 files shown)


app/mailers/sync_mailer.rb
```diff
@@ -1,63 +1,15 @@
 class SyncMailer < ApplicationMailer
-  def sync_notification(email:, integration_id:, results:)
-    @integration_id = integration_id
-    @sync_results = SyncResults.new(integration_id, results)
+  def sync_notification(email:, integration_id:, sync_summary:)
```
derekprior: sync_summary has a connection, which has an integration_id, so you could eliminate that parameter, I think?

derekprior:
:+1:

app/mailers/sync_mailer.rb
jferris: Done in ef32553.

PR #204, opened 2015-09-03
Display authorization errors in activity feed

derekprior:
Previously, the only user-facing notification of authorization errors
was in an email notification. Adding those events to the activity feed
makes it so there's a centralized place to see what's happening with
your integration.


derekprior:
namely_connect

jferris:
This looks good to me.

PR #205, opened 2015-09-03
Fix NetSuite employee creation

jferris:
The Namely API actually return empty string rather than null for new
employees.

This has always been the case, but a recent refactoring changed it to
check for null (which we were using in our fixtures) rather than empty
string.

This changes the fixtures to be more realistic, so that we'll catch
future regressions, and fixes the NetSuite client to create a new
employee when it receives an empty string ID.

### DIFF
--- a/spec/models/net_suite/export_spec.rb
+++ b/spec/models/net_suite/export_spec.rb
@@ -4,21 +4,17 @@ describe NetSuite::Export do
   describe "#perform" do
     context "with new, valid employees" do
       it "returns a result with created profiles" do
-        emails = %w(one@example.com two@example.com)
         profile_data = [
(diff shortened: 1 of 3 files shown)


spec/models/net_suite/export_spec.rb
```diff
@@ -91,8 +85,8 @@ def build_namely_connection
     create(:user).namely_connection
   end
 
-  def stub_profile(overrides = {}, data = {})
-    data.stringify_keys.tap do |profile|
+  def stub_profile(overrides = {})
```
jferris: Addressed in 2c00e83.

derekprior:
:+1:

PR #206, opened 2015-09-03
Send all unauthorized emails from notifier

derekprior:
Previously, the unauthorized email was being sent via model method on
`Installation`. The `UnauthorizedNotifier` used this as did a couple of
places in other code.

I removed the ActiveRecord model methods concerned with mail delivery
and opted to send the mail directly via the `UnauthorizedNotifier` just
as we do with `SyncJob`. Then all places that were sending the mail
directly, without the notifier, were updated to use the notifier.

This centralizes the responsibility for tracking unauthorized sync
summaries and ensures we get both an email and a summary object created.

### DIFF
--- a/app/mailers/connection_mailer.rb
+++ b/app/mailers/connection_mailer.rb
@@ -1,7 +1,7 @@
 class ConnectionMailer < ApplicationMailer
-  def authentication_notification(integration_id:, email:, message:)
-    @integration = map_integration_id_to_name(integration_id)
-    @message = message
+  def authentication_notification(email:, sync_summary:)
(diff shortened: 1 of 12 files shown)


jferris:
Nice, this looks good!

This will also create activity feed entries, right? (Even though there isn't a link to them yet)

derekprior:
Yes, this also creates activity feed entries. Which means that for these integrations, the only thing that would ever be in them were these autheorization errors. But we don't expose them yet anyway.

derekprior:
Merged in 30fe8dfc01

PR #207, opened 2015-09-08
Use reverse chronological order in activity feed

derekprior:


spec/models/sync_summary_spec.rb
```diff
@@ -16,4 +16,13 @@
       expect(summary.profile_events.pluck(:profile_name)).to match_array(names)
     end
   end
+
+  describe ".ordered" do
+    it "orders by created_at, descending" do
```
jferris: Does this test pass if you use `SyncSummary.all` instead of `ordered`? I usually set three instances out of order with explicit `created_at` values, because the test passes without implementing it otherwise.

derekprior: No, it failed with that implementation.

jferris: Okay. I think Postgres order is non-deterministic without an `ORDER BY` clause, but I also think the risk of a regression here is not very high.

jferris:
Looks good to merge.

PR #208, opened 2015-09-08
Add feature flag for NetSuite custom fields

derekprior:
We have long-standing issue with Cloud Elements concerning our inability
to set NetSuite custom fields via their API. We have opted to remove
custom fields from the first phase of the project. I have kept the
functionality behind a feature flag, which allows us to continue to test
it locally as we get new information from Cloud Elements.


app/models/net_suite/normalizer.rb
```diff
@@ -36,8 +36,11 @@ def mapped_attributes
         "gender" => gender,
         "isInactive" => user_status,
         "subsidiary" => subsidiary,
-        "customFieldList" => custom_fields
-      }
+      }.tap do |attributes|
```
jferris: In `to_hash`, we merge a series of hashes together to calculate the final hash. What do you think about extracting out a `custom_fields_attributes` method with this logic and merging it in `to_hash` with the others?

derekprior: good idea. Ready for another look.

jferris:
Looks good to merge!

PR #209, opened 2015-09-08
Handle more Errors from NetSuite

derekprior:
We occasionally get errors back from NetSuite other than 400 (Bad
Request) or 401 (Unauthorized). When this happense, we're crashing the
entire job.

By handling `RestClient::Exception`, we can handle *all*  unexpected
HTTP response codes, allowing the job to continue running to completion
and record a proper `SyncSummary`.

### DIFF
--- a/app/models/net_suite/export.rb
+++ b/app/models/net_suite/export.rb
@@ -80,7 +80,7 @@ module NetSuite
       rescue NetSuite::ApiError => exception
         Result.new(
           success: false,
-          error: exception.message,
+          error: exception.to_s,
           updated: updated,
           profile: @profile
         )
--- a/app/models/net_suite/api_error.rb
+++ b/app/models/net_suite/api_error.rb
@@ -1,15 +1,22 @@
 module NetSuite
-  # Wraps BadRequest errors from Cloud Elements. Attempts to parse a
+  # Wraps HTTP errors from Cloud Elements. Attempts to parse a
   # human-readable error message from the response.
   class ApiError < StandardError
-    def initialize(response)
-      @response = response
+    attr_reader :http_code
+
+    def initialize(exception)
+      @response = exception.response
+      @http_code = exception.http_code
     end
 
     def message
       response_data["providerMessage"] ||
         response_data["message"] ||
         "Unknown error"
+    end
+
+    def to_s
+      "NetSuite API Error: #{http_code} - #{message}"
     end
 
     private
--- a/app/models/net_suite/client/request.rb
+++ b/app/models/net_suite/client/request.rb
@@ -38,11 +38,13 @@ module NetSuite
         if response.present?
           JSON.parse(response)
         end
-      rescue RestClient::BadRequest => exception
-        Rails.logger.error "Bad Request: #{exception.response}"
-        raise NetSuite::ApiError, exception.response
       rescue RestClient::Unauthorized => exception
         raise Unauthorized, exception.message
+      rescue RestClient::Exception => exception
+        api_error = NetSuite::ApiError.new(exception)
+        Raygun.track_exception(api_error)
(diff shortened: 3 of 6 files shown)


jferris:
Much better. Do we need to explicitly deliver these errors to Raygun, or do we assume the Raygun client is doing that for us?

derekprior:
We're now handling all RestClient based response code exceptions, and thus not raising them. They will be reported to the user but not logged to Raygun. We could potentially inspect the HTTP status code and report anything that's 500+, but I don't know that we need to do that. Perhaps best to monitor the `profile_events` table itself for that kind of data.

I have another ticket I'm investigating now that set DJ/Raygun to report any errors that do get raised from a delayed job so at least we're alerted rather than having silent failures. I think I'll end up having to do something like this: https://github.com/honeybadger-io/delayed_job_honeybadger/blob/master/lib/delayed/plugins/honeybadger.rb

jferris:
I think we probably want a developer-visible log of what error messages are coming back from Cloud Elements. Some of them are definitely user error (like bad passwords), but we don't know what other kinds of errors are lurking out there, as we've learned. For example, if we display "Method not allowed" to the user without notifying developers, we'll displaying an opaque message to the user and not noticing a bug.

derekprior:
What if we increased the specificity/reportability of the profile_events table to list the http_code as a separate queryable field?

I'm worried about handling it in the code that handles `RestClient::Exception` and turns it into `ApiError` for a couple of reasons:

1. It makes the assumption in that location that `ApiError` is eventually going to be caught somewhere. Assumptions that are true today, but not necessarily tomorrow.

2. It's very likely to over-report certain errors. We could filter out 400 and 401 specifically, I guess?

derekprior:
Oh, I guess I could solve conern #1 by logging it in Raygun here instead: https://github.com/namely/connect/blob/master/app/models/net_suite/export.rb#L80

Then it's just a matter of limiting it to actionable errors. Would you suggest waiting and seeing on that?

jferris:
> What if we increased the specificity/reportability of the profile_events table to list the http_code as a separate queryable field?

I think that couldn't hurt. Might be good to store the full response body as well.

> It makes the assumption in that location that `ApiError` is eventually going to be caught somewhere. Assumptions that are true today, but not necessarily tomorrow.

I'm not sure I understand this. We currently rescue it in a few places, but we generally expect it to render a 500 error or end the export for the current profile. That seems like what we'd want for all API errors.

> It's very likely to over-report certain errors. We could filter out 400 and 401 specifically, I guess?

I think it's probably worth worrying about that later. Early on, I think it's better to get too much feedback than too little.

derekprior:
I'm now logging all API errors during a sync to Raygun. I punted on updating the profile events for now. That may be a better soltuion in the longterm, but it will be interesting to see the volume first.

app/models/net_suite/export.rb
```diff
@@ -78,9 +78,11 @@ def request(updated:)
           profile: @profile
         )
       rescue NetSuite::ApiError => exception
+        Raygun.track_exception(exception)
```
jferris: Is there a reason to track it here rather than in `NetSuite::Client::Request`? This will only track export exceptions, whereas tracking it in `Request` would also track exceptions during instance creation, subsidiary selection, and field mapping.

derekprior: I liked that it left the code here responsible for translating exceptions and left what to do about those exceptions up to the calling code.

Are we really going to want a raygun error every time someone puts their password or account Id incorrect when connecting?

derekprior: I'll move the code that does the logging into the t`Request` and then see how it goes.

jferris:
I think this is ready to merge whenever you are.

PR #210, opened 2015-09-09
Report DelayedJob exceptions to Raygun

derekprior:
Raygun doesn't ship with support for DelayedJob out of the box, so we
have to use DelayedJob's plugin system to catch and report these. I
based this plugin off of this: http://blog.salsify.com/engineering/delayed-jobs-callbacks-and-hooks-in-rails

I could not conceive of a good way to test this in an automated fashion,
so I tested it in development by setting SyncJob to raise from `perform`
and making sure the `rescue` block in the plugin was executing and that
DelayedJob still requeued the job for a retry.

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,0 +1,3 @@
+require "delayed_raygun"
+
+Delayed::Worker.plugins << DelayedRaygun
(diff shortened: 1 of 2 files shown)


jferris:
This looks good to me.

Hopefully ActiveJob will unify some of this stuff sometime soon.

PR #211, opened 2015-09-09
Limit DelayedJob retries

derekprior:
By default, DelayedJob will retry a job 25 times before failing. If it
goes that long, the last retry will be 20 days after the job was first
enqueued, after a 100 hour backoff interval, due to the exponential
backoff.

The jobs we run do not need to hang around that long. The NetSuite sync
job will be re-run daily anyway. Setting the max retries to 4 means that
the final backoff interval is just under 5 minutes. All told it would
take under 10 minutes for us to fail a job entirely.

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,3 +1,4 @@
 require "delayed_raygun"
 
+Delayed::Worker.max_attempts = 4
 Delayed::Worker.plugins << DelayedRaygun


jferris:
I think this is probably a good idea.

We may want to avoid retrying the sync jobs at all. We have no reason to believe a retry will work, and users will eventually have the ability to manually trigger a retry through the UI.

derekprior:
Unfortunately, ActiveJob does not allow overriding max retries on a per-job basis when DJ is the backing queue.

* http://edgeapi.rubyonrails.org/classes/ActiveJob/QueueAdapters.html
* https://github.com/collectiveidea/delayed_job/issues/715

:frowny_face: :sad_times:

I think this is a good enough compromise for now.

jferris:
> Unfortunately, ActiveJob does not allow overriding max retries on a per-job basis when DJ is the backing queue.

Sorry, I meant we could just rescue the `ApiError` and not re-raise it, since we've already been notified of it and we don't want the job to fail/retry.

> I think this is a good enough compromise for now.

Agreed.

derekprior:
> Sorry, I meant we could just rescue the ApiError and not re-raise it, since we've already been notified of it and we don't want the job to fail/retry.

That does happen. `ApiError` gets turned into a failed result. It'd be other exceptions (unforseen, I guess) that would trigger a retry now.

PR #212, opened 2015-09-09
Add background job to refresh Namely access tokens

derekprior:
We've seen situations where users that own an integration end up having
their access tokens go stale, which causes 401 errors. In order to
prevent users from having stale access tokens, we are creating a
background job to be run daily that refreshes all tokens.

### DIFF
--- a/app/jobs/token_refresh_job.rb
+++ b/app/jobs/token_refresh_job.rb
@@ -1,0 +1,5 @@
+class TokenRefreshJob < ActiveJob::Base
+  def perform(user)
+    user.fresh_access_token
+  end
+end
--- a/lib/tasks/namely.rake
+++ b/lib/tasks/namely.rake
@@ -1,0 +1,8 @@
+namespace :namely do
+  desc "Refresh access tokens for all users"
+  task refresh_access_tokens: :environment do
+    User.all.each do |user|
+      TokenRefreshJob.perform_later(user)
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


jferris:
We know that this won't work for users with expired refresh tokens, right? We may want to filter those users out. Otherwise, we'll be creating background jobs for them every time this task runs, and each job will fail and retry four times.

derekprior:
Refresh tokens don't have (or store?) an expiration. Only the access token stores an expiration. The access token itself is never refreshed unless the previous one expired, so I could eliminate some work by only selecting users that have expired access tokens.

If refresh tokens don't expire, maybe the better solution is to check access token expiration before any namely request and refresh it if necessary.

It feels like I'm missing something on how the tokens work.

derekprior:
Some investigation results:

1. Whenever a namely connection is used, we use the user's refresh token to get a new access token if necessary.

2. Getting a new access token does not seem to invalidate the previous refresh token. It also does not return a new one.

3. I don't see any methods in the Namely api client that would let me refresh a refresh token.

4. According to Attila, refresh tokens don't expire for a year, so this shouldn't be a pressing concern. However, Namely may revoke the tokens. We should probably handle this, I'm just not sure how at this moment. There's no way to get a new refresh token without going through the connection process again.

Given the above, I'm going to close this PR. I think we should instead focus on how to alert the user that this is a problem.

PR #215, opened 2015-09-16
Improve activity feed performance

derekprior:
* Move scopes to associations to allow them to be eager loaded.
* Page activity feed


derekprior:
I'm not sure what hound is complaining about here. This is how we continue long lines, right?

jferris:
This looks good to me. I'm unsure of what Hound is asking for.

PR #216, opened 2015-09-16
Allow failed profiles to be retried

derekprior:
An installation may have hundreds or even thousands of namely profiles
with only a small number of failures. In these cases, it's reasonable
for users to want to quickly retry only the failed profiles once they
think they have made sufficient corrections.

To facilitate, this change introduces a `RetryJob` which will sync only
the failed profiles associated with a sync summary. Syncing was built to
sync all namely profiles over a given connection, so we needed a way to
tell the sync operation to only sync a subset of profiles. An optional
named argument of `profiles` was added to the net suite sync.

A `Sync` class was also extracted to handle the common concern of
notifications between the `SyncJob` and `RetryJob`. This class will pass
on any keyword arguments it receives to the sync operation. To avoid
having to do a conditional on the presence of keyword arguments, the
Jobvite `sync` definition had to be updated to accept keyword arguments
because it will be passed an empty hash of them.

### DIFF
--- a/app/assets/stylesheets/_activity-feed.scss
+++ b/app/assets/stylesheets/_activity-feed.scss
@@ -22,4 +22,13 @@
     color: $grey-75;
     font-weight: bold;
   }
+
+  .button_to {
+    float: right;
+  }
+
+  input {
+    @extend .-n-btn;
+    @extend .-n-btn-sm;
+  }
 }
--- a/app/controllers/retries_controller.rb
+++ b/app/controllers/retries_controller.rb
@@ -1,0 +1,21 @@
+class RetriesController < ApplicationController
+  def create
+    sync_summary = find_sync_summary
+    RetryJob.perform_later(sync_summary)
+
+    redirect_to sync_summary_retry_path(sync_summary: sync_summary)
+  end
+
+  def show
+  end
+
+  private
+
+  def find_sync_summary
+    SyncSummary.find(params[:sync_summary_id]).tap do |summary|
+      unless summary.users.include?(current_user)
+        raise ActiveRecord::RecordNotFound
+      end
+    end
+  end
+end
--- a/app/jobs/retry_job.rb
+++ b/app/jobs/retry_job.rb
@@ -1,0 +1,7 @@
+class RetryJob < ActiveJob::Base
+  def perform(sync_summary)
+    Notifier.execute(sync_summary.connection) do
+      sync_summary.retry
+    end
+  end
+end
--- a/app/models/jobvite/connection.rb
+++ b/app/models/jobvite/connection.rb
@@ -87,6 +87,10 @@ module Jobvite
       delegate :name, to: :candidate
       delegate :success?, :error, to: :result
 
+      def profile_id
+        ""
+      end
+
       private
 
       attr_reader :candidate, :result
--- a/app/models/sync_summary.rb
+++ b/app/models/sync_summary.rb
@@ -12,7 +12,9 @@ class SyncSummary < ActiveRecord::Base
     class_name: "ProfileEvent"
   )
 
-  delegate :integration_id, to: :connection
+  delegate :installation, :integration_id, to: :connection
+  delegate :users, to: :installation
+  delegate :namely_profiles, to: :installation
 
   def self.create_from_results!(results:, connection:)
     transaction do
@@ -30,5 +32,17 @@ class SyncSummary < ActiveRecord::Base
     results.each do |result|
       profile_events.create_from_result!(result)
     end
+  end
+
+  def failed_profiles
+    failed_ids = failed_profile_events.pluck(:profile_id)
+
+    namely_profiles.select do |profile|
+      failed_ids.include?(profile.id)
+    end
+  end
+
+  def retry
+    connection.retry(self)
   end
 end
--- a/db/migrate/20150903174313_add_profile_id_to_profile_events.rb
+++ b/db/migrate/20150903174313_add_profile_id_to_profile_events.rb
@@ -1,0 +1,10 @@
+class AddProfileIdToProfileEvents < ActiveRecord::Migration
+  def up
+    add_column :profile_events, :profile_id, :string
+    change_column_null :profile_events, :profile_id, false, ""
+  end
+
+  def down
+    remove_column :profile_events, :profile_id
+  end
+end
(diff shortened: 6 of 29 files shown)


app/models/jobvite/connection.rb
```diff
@@ -48,7 +48,7 @@ def required_namely_field
       normalizer.namely_identifier_field.to_s
     end
 
-    def sync
+    def sync(_: nil)
```
derekprior: This seems really screwy, but it allows the `Sync` service object to call `connection.sync(args)` where `args` may be an empty hash of keyword arguments. Too clever, maybe?

Other thing I considered was a conditional in `Sync` that would call `connection.sync` with no arguments if there were no keyword args (rather than an empty hash).

Another alternative, which I ruled out as too drastic was to change syncing to occur on the profile level (taking a connection) rather than on the connection level.

app/controllers/retries_controller.rb
```diff
@@ -0,0 +1,11 @@
+class RetriesController < ApplicationController
+  def create
+    sync_summary = SyncSummary.find(params[:sync_summary_id])
```
jferris: Not much of a vulnerability, but do we want to verify that the user owns the connection the sync summary refers to?

app/jobs/retry_job.rb
```diff
@@ -0,0 +1,17 @@
+class RetryJob < ActiveJob::Base
+  def perform(sync_summary)
+    connection = sync_summary.connection
+    profiles = failed_profiles(sync_summary)
+    Sync.perform(connection, profiles: profiles)
+  end
+
+  private
+
+  def failed_profiles(sync_summary)
```
jferris: Since this only uses methods on `sync_summary`, does it make sense to move this method to that class?

app/models/jobvite/connection.rb
jferris: What do you think about accepting a `previous_summary` parameter or something similar which responds to `unsynced_profiels`?

When doing a fresh sync, we can pass a null object which returns all profiles. When doing a retry, we can pass the previous `SyncSummary`, which would return only the failed profiles.

app/models/sync_summary.rb
```diff
@@ -31,4 +31,8 @@ def convert_results_to_profile_events(results)
       profile_events.create_from_result!(result)
     end
   end
+
+  def installation
```
jferris: It looks like this is only used to provide `namely_profiles`. Does it make more sense to expose that directly instead?

db/migrate/20150903174313_add_profile_id_to_profile_events.rb
```diff
@@ -0,0 +1,9 @@
+class AddProfileIdToProfileEvents < ActiveRecord::Migration
+  def up
+    add_column :profile_events, :profile_id, :string
```
jferris: Is this going to be an issue for existing profile events?

Should this column actually be `NOT NULL`?

app/controllers/retries_controller.rb
derekprior: Fixed in 6b64677

app/models/jobvite/connection.rb
derekprior: The problem with that is that jobvite doesn't care about your namely profiles at all. It wouldn't use any parameter that we passed into it because it's syncing users in the other direction.

The fact that both `NetSuite::Connection` and `Jobvite::Connection` have a `sync` method seems more a happy accident of naming than anything else. They have wildly different reasons to change and this issue is a symptom of that.

app/jobs/retry_job.rb
derekprior: Done in deed10f08635da51cf327820a47da2415368fc58

app/models/sync_summary.rb
derekprior: done in deed10f08635da51cf327820a47da2415368fc58

db/migrate/20150903174313_add_profile_id_to_profile_events.rb
derekprior: Fixed in 99f4153.

I made it `null: false` and updated jobvite to persist an empty string there.

app/models/jobvite/connection.rb
jferris: I made all the integrations use `sync` during an effort to bring them all closer in terms of implementation, since there has historically been a lot of duplication.

We do want to support retry for other integrations like Jobvite. In the import case, we'd want to retry profiles which failed to import, so we'd ask for Jobvite candidates instead.

I think it's fine to ignore the argument for now in the Jobvite connection. I think the reason this is so screwy is that the integrations haven't completely evolved together (for example, they still don't all support field mappings), so we have some warts like this and the `has_activity_feed?` style query methods.

I suggested passing that object in because I thought it would help clean up the NetSuite implementation, which currently has a control couple based on the parameter.

derekprior: Reworked as we discussed in 75f5ea5

app/models/jobvite/connection.rb
```diff
@@ -52,6 +52,10 @@ def sync
       import_results.map { |result| Result.new(**result) }
     end
 
+    def namely_profiles
```
jferris: Did this end up being used?

app/assets/stylesheets/_activity-feed.scss
```diff
@@ -22,4 +22,13 @@
     color: $grey-75;
     font-weight: bold;
   }
+
+  .button_to {
```
derekprior: We're styling a class rails gives us in this case.

db/migrate/20150903174313_add_profile_id_to_profile_events.rb
```diff
@@ -0,0 +1,11 @@
...
+    execute "UPDATE profile_events SET profile_id = '' WHERE profile_id IS NULL"
```
jferris: I think you can do this by providing `""` as the fourth argument to `change_column_null`.

jferris:
I had a couple of minor suggestions, but I think this is ready to merge whenever you're feeling good about it.

app/models/jobvite/connection.rb
derekprior: nope. Meant to remove it

PR #217, opened 2015-09-22
Include custom fields on greenhouse imports

derekprior:
Custom fields are supported in Greenhouse imports so long as there is an
identically named namely field of the proper type. Unfortunately, when
Greenhouse was moved to the attribute mapper, we stopped syncing these
custom fields.

The custom fields didn't have attribute mappings (and can't, because we
have no way to get a list of fields from Greenhouse until they send us a
user), so the attribute mapper was dropping them. To fix this, we have
the normalizer specifically whitelist any of the matching custom fields.

### DIFF
--- a/app/models/greenhouse/normalizer.rb
+++ b/app/models/greenhouse/normalizer.rb
@@ -8,10 +8,16 @@ module Greenhouse
     end
 
     def call(data)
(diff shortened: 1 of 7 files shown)


jferris:
This looks good to me.

PR #223, opened 2015-09-29
Change status to 202 for unauthorized webhooks to prevent retries

bobbytables:
When we return a non-200 status code, greenhouse keeps sending us data, which causes several error emails to go out to customers. This changes it to a 202 accepted status to indicate we recognize the request.

See:
gmail_-_namely


naush:
:+1:

cored:
:+1: @bobbytables just fix the things that hound is saying and you will be good to go ;-)

PR #229, opened 2015-10-09
Extra Netsuite parameters for SSO mapping

macat:
In order to authenticate the user properly with NetSuite, we need to set a couple of extra fields, like `roleId`, `appId`.

This PR adds these to the cloud-elements instance creation.

There's a subtle modification on the NEtsuite Credentials page.

### DIFF
--- a/app/models/net_suite/authentication.rb
+++ b/app/models/net_suite/authentication.rb
@@ -1,6 +1,7 @@
 module NetSuite
   class Authentication
     include ActiveModel::Model
+    extend Forwardable
 
     attr_accessor :account_id, :email, :password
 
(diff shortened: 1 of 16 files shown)


config/locales/en.yml
```diff
@@ -179,6 +179,10 @@ en:
     description:
       connected_html: Connected NetSuite Description
       disconnected_html: Disconnected NetSuite Description
+    authentication_form:
+      help: >
+        Please note that the credentials are needed for initial connection to
```
bobbytables: I think this might be clearer if we change it to "Please note that these credentials ...". Since it's displayed inline with them on the form.

app/models/net_suite/authentication.rb
```diff
@@ -22,6 +23,22 @@ def update(attributes)
       valid? && create_instance
     end
 
+    def user_id
+      @connection.installation.owner.namely_user_id
```
bobbytables: Law of demeter.

bobbytables:
Looking good. A few comments.

macat:
Fixed 'em.

PR #238, opened 2015-11-05
Netsuite Matcher

macat:
Netsuite Matcher will be used to find matching employee objects between netsuite and namely.

@bobbytables


macat:
+1 from @bobbytables

PR #242, opened 2015-11-10
Allow switching matching strategy

bobbytables:
This is so clients can choose which matching strategy they'd like to use for NetSuite exports.


macat:
Why don't we just default this to email, and add the option when clients request it?

macat:
LGTM +1

PR #243, opened 2015-11-11
Netsuite job

macat:


app/jobs/net_suite_export_job.rb
```diff
@@ -0,0 +1,48 @@
...
+             end
+
+    ProfileEvent.create(
+      sync_summary_id: sync_summary_id,
+      profile_id: result.profile_id,
+      profile_name: result.name,
+      error: result.error
+    )
+  end
+end
```
bobbytables: As long as this method is, I really like this. It does everything we need it to do.

spec/features/user_retries_failed_profile_sync_spec.rb
```diff
@@ -14,6 +15,8 @@
     click_on t("profile_event.retry")
     visit_activity_feed(connection, user)
 
+    save_page
```
bobbytables: This needs to go away.

bobbytables:
:shipit: @macat

PR #258, opened 2015-11-23
Only pull employees from netsuite for the subsidiary that we care about.

bobbytables:
This change makes sure that we only pull the employees for the subsidiary set on the initial connection setup for netsuite. Since companies could have many subsidiaries, this is important to handle.


macat:
+1

PR #261, opened 2015-11-24
Update jobvite normalizer to include jobvite_id

bobbytables:
This fixes a problem where the normalizer would never actually output the jobvite_id. Because we stub everything, we stubbed the wrong thing and didn't catch this. But the truth is that the attribute mapper never actually maps the jobvite_id, so it would never be exported. This just merges in the jobvite_id attribute.

/cc @macat


macat:
+1

PR #263, opened 2015-11-30
Add hireDate as a standard field for netsuite.

bobbytables:
We didn't have `hireDate` as a standard Netsuite field in our attribute mapper. This PR adds it as a standard field.

This PR also boyscouts a lot of issues with the Namely API changing under our feet, for example:

* Date formats changing causing date parsing to fail
* Profile JSON being shorted. (Fixed with a monkey patch)

### DIFF
--- a/app/models/net_suite/normalizer.rb
+++ b/app/models/net_suite/normalizer.rb
@@ -40,6 +40,7 @@ class NetSuite::Normalizer
         "isInactive" => user_status,
         "subsidiary" => subsidiary,
         "releaseDate" => release_date,
+        "hireDate" => hire_date,
       }
     end
 
@@ -73,6 +74,14 @@ class NetSuite::Normalizer
 
       if date.present?
         date.to_datetime.to_i * 1.second.in_milliseconds
+      end
+    end
+
+    def hire_date
+      date = @attributes.fetch("hireDate", Fields::NullValue.new).to_date
+
+      if date.present?
+        date.to_datetime.iso8601
       end
     end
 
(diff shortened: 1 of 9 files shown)


app/models/net_suite/normalizer.rb
```diff
@@ -76,6 +77,14 @@ def release_date
       end
     end
 
+    def hire_date
+      date = @attributes.fetch("hireDate", Fields::NullValue.new).to_date
```
macat: Does `@attributes.fetch` always return something with `to_date`?

bobbytables: Well the logic here is that if the key doesn't exist, return the NullValue which has a method `to_date` that returns nil.

Otherwise, this value will always be a string from Netsuite, which won't fail (thanks to Rails)

macat: perfect.

config/initializers/namely_client.rb
```diff
@@ -0,0 +1,20 @@
+Namely::ResourceGateway # Poke the constant to make sure we're not defining from scratch
+
+module Namely
+  class ResourceGateway
+    def json_index_paged
```
macat: gotta love monkey patching! :dancer:

bobbytables: Ayyyyyyy. haha.

config/initializers/namely_client.rb
```diff
@@ -0,0 +1,20 @@
...
+      Enumerator.new do |y|
+        params = { profile_format: 'full' }
```
bobbytables: This is really the only thing that was needed to be patched. :p

macat: that is sad :cry:

macat:
LGTM

PR #266, opened 2016-01-11
Add NetSuite::ProfilesSorter

cored:
We need a way to create managers before the profiles that reports to
them.


Gemfile.lock
```diff
@@ -361,6 +361,3 @@ DEPENDENCIES
   title
   unicorn
   webmock
-
-BUNDLED WITH
-   1.10.6
```
bobbytables: I think we've mentioned this at least twice now, but _please_ run `gem update bundler`

app/services/net_suite/profiles_sorter.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      attr_reader :guid
+      def initialize(guid:)
+        @guid = guid
+      end
+
+      def <=>(other_guid)
+        guid <=> other_guid
+      end
```
bobbytables: So, the problem I see with this implementation is that uuid's are not sortable. They aren't a sequential sequence.

Gemfile.lock
cored: @bobbytables ok, updating bundler

app/services/net_suite/profiles_sorter.rb
cored: @bobbytables will use proper uuid in the specs to test this out.

Gemfile.lock
bobbytables: Re-bundling should add this back in.

bobbytables:
:shipit: :+1:

cored:
@bobbytables Reopening this because of the revert.

bobbytables:
Thank you. Sorry for the git history mess I made.

PR #267, opened 2016-01-19
Rescue phone number formatting

bobbytables:


app/services/net_suite/diff_normalizer.rb
```diff
@@ -27,6 +27,10 @@ def normalize_phone_numbers
       number = GlobalPhone.parse(employee[phone_key])
       employee[phone_key] = number.national_format
     end
+  rescue StandardError => e
```
cored: @bobbytables is there a reason on why you did not put the specific thrown exception in here? Just curious?

PR #270, opened 2016-01-25
Revert "Revert "Add NetSuite::ProfilesSorter""

bobbytables:
Reverts namely/connect#268

REVERT-CEPTION. :fire: :fire: :fire: :fire:


app/services/net_suite/profiles_sorter.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def call
+      profiles.reduce([]) do |sorted_profiles, profile|
+        report_to = ReportTo.for(profile)
+        sorted_profiles.take_while do |p|
+          p.id != report_to.id
+        end + [profile] + sorted_profiles.drop_while do |p|
+          p.id != report_to.id
+        end
```
bobbytables: Here's where we'll need to append the `supervisor_netsuite_id` to the profile.

app/services/net_suite/profiles_sorter.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    class ProfileWithSupervisorId < SimpleDelegator
+      def initialize(profile:, reports_to:)
+        @reports_to = reports_to
+        super(profile)
+      end
+
+      def class
+        __getobj__.class
+      end
```
bobbytables: Interesting. Why is this here?

cored: @bobbytables It just osd on my part. So according to the gang of four the decorator should have be the same class as the underlying object that you want to expand the public API upon; that's why that is there. But right now is not needed.

bobbytables: I think if we're going to introduce code like this we should at a minimum document the method with a comment explaining it's purpose.

But I'm also not the biggest fan of not introducing code that isn't used :/

cored: You know what, I will remove this. It doesn't give any value.

PR #271, opened 2016-01-25
Feature: Add mapping for Reports to from NetSuite to Namely in User Experience

cored:
Fix: https://namely.atlassian.net/browse/CORE-88


app/models/net_suite/normalizer.rb
```diff
@@ -89,6 +90,12 @@ def subsidiary
       { "internalId" => @subsidiary_id }
     end
 
+    def supervisor
+      {
+        "internalId" => @attributes.fetch("internalId", Fields::NullValue.new).to_s
```
bobbytables: So this logic isn't quite right I think. It needs to be the internalId of the supervisor, not the current profile being exported.

cored: I see, I'll check thst out.

spec/models/net_suite/normalizer_spec.rb
```diff
@@ -24,7 +26,8 @@
         subsidiary
         releaseDate
         nullFieldList
-      ))
+        supervisor
+                                                    ))
```
poopoothegorilla: this looks wierd haha

bobbytables: I agree.

spec/models/net_suite/normalizer_spec.rb
```diff
@@ -178,6 +181,25 @@
...
+        ).to eq("internalId" => "456")
+      end
+
+      it "returns a blank without a netsuite_id" do
+        export_attributes = export
+
+        expect(
+          export_attributes["supervisor"]
+        ).to eq("internalId" => nil)
+      end
```
bobbytables: We may need to validate if this breaks when we send the data over to NetSuite.

bobbytables: The nil value to be specific.

cored: @bobbytables got it

PR #272, opened 2016-01-27
Can't load Gemfile when loading the app through docker

cored:
Fix: CORE-115
https://namely.atlassian.net/browse/CORE-115


bobbytables:
I'm not sure this is a bug. What problem are you experiencing?

cored:
@bobbytables https://github.com/docker/compose/issues/2103

PR #276, opened 2016-02-08
Respond 200 to greenhouse secret keys we don't know about.

bobbytables:
This is because clients remove the connection from Namely Connect for greenhouse but don't disable the webhook from their Greenhouse account. So we still get webhooks dispatched to us. If we don't respond with an OK status message they'll keep dispatching to us as a retry forever. This mitigates our error capturing getting blown up.


macat:
+1

poopoothegorilla:
+1

PR #283, opened 2016-02-25
CORE-114: Disable supervisor mapping in ui

cored:
- Add initial code for assigning a supervisor to the imported profile
- Add supervisor field to get mapped on import Using internalId as the supervisor field
- Refactor the code related to supervisor to use a new field generated by the ProfileSorter
- Fix failing spec after adding new mapping for NetSuite::Connection
- Rename supervisor_netsuite_id to netsuite_supervisor_id
- Add NamelyFieldsByLabel decorator class to add the supervisor netsuite id field
- Add NamelyFieldsByLabel#disable_namely_fields to disable every other field so the user can't not map it
- Add implementation for #disable_namely_fields
- Add assertion to check for the Supervisor dropdown in the mapping for Netsuite
- Remove expectation for User#namely_fields_by_label the logic is abstracted in NamelyFieldsByLabel
- Add initial code for assigning a supervisor to the imported profile
- Add supervisor field to get mapped on import
- Using internalId as the supervisor field
- Add NamelyFieldsByLabel decorator class to add the supervisor netsuite id field
- Add NamelyFieldsByLabel#disable_namely_fields to disable every other field so the user can't not map it
- Remove the disabling of the supervisor id, need to think in a better way to implement this right now disabling one field in the drop down disables all the others
- Add proper assertion to disabling drop down for supervisor netsuite id
- Add #is_disable_field? to check if the current namely field is marked for disabling

### DIFF
--- a/app/models/namely_fields_by_label.rb
+++ b/app/models/namely_fields_by_label.rb
@@ -1,0 +1,26 @@
+class NamelyFieldsByLabel
+  DISABLE_FIELDS = [
+    ["Supervisor NetSuite Id", "netsuite_supervisor_id"]
+  ]
+  def initialize(namely_fields:)
+    @namely_fields = namely_fields
+  end
+
+  def to_a
+    namely_fields + DISABLE_FIELDS
+  end
+
+  def is_disable_field?(namely_field_name)
+    DISABLE_FIELDS.map do |disable_field|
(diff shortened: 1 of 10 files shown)


bobbytables:
:shipit:

macat:
+1

PR #287, opened 2016-04-12
Our emails make it sound like it was the integrated partners fault when it usually isn't.

bobbytables:


config/locales/en.yml
```diff
@@ -102,7 +102,8 @@ en:
         Reason:
         %{status_error}
 
-        Please fix these errors with %{integration} and retry the import.
+        Please fix these errors and retry the import. If you need more help,
```
cored: :-)

therealplato:
+1

PR #301, opened 2016-05-18
Bellis core 287 (#300)

ellisbenjamin:
* Disconnecting from a integration now flashes successful messages on success.


ellisbenjamin:
As a user I want to be notified when I disconnect an integration via flash message so I am confident that the disconnection was successful.

PR #305, opened 2016-05-24
Dropdown boxes now look like drowndown boxes rather than ambiguous in…

ellisbenjamin:
Dropdown boxes now look like drowndown boxes rather than ambiguous input boxes

https://namely.atlassian.net/browse/CORE-291


bobbytables:
"drowndown" lol

PR #312, opened 2016-05-26
Bellis core 103

ellisbenjamin:
As Namely Connect, I need to capture information about clients that are setup to use Namely Connect so that I am able to know all clients using the Connectors.

https://namely.atlassian.net/browse/CORE-103


bobbytables:
👍 👍 👍 👍

PR #313, opened 2016-05-31
Bellis use owner id for owner

ellisbenjamin:
As a user of namely-connect I should be able to change the owner of the installation via the view all users page.

https://namely.atlassian.net/browse/CORE-302


app/controllers/users_controller.rb
```diff
@@ -1,5 +1,12 @@
 class UsersController < ApplicationController
   def index
+    @installation_id = current_user.installation.id
     @users = current_user.installation.users
+    @owner_id = current_user.installation.owner.id
+  end
+
+  def update
+    current_user.installation.update(owner_id: params[:user_radio])
```
bobbytables: There's a small security concern here. There's a possibility of changing the installations owner_id to an ID that someone has posed as.

Proposed solution:

Find the User in the database for the installation before updating this attribute. So:

```ruby
def update
  installation = current_user.installation
  if user = installation.users.find(params[:user_radio])
    current_user.installation.update(owner_id: user.id)
  end
end
```

app/views/users/index.html.erb
```diff
@@ -1,18 +1,24 @@
...
+      <tr>
+        <th>Owner?</th>
+        <th>First name</th>
+        <th>Last name</th>
+        <th>Email</th>
+      </tr>
+        <% @users.each do |user| %>
+          <tr>
+              <td><%= radio_button_tag("user_radio", user.id, user.id == @owner_id) %>
+              </td>
```
bobbytables: Formatting.

spec/features/user_views_list_of_users_and_owners_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    expect(find("#user_radio_#{users[0].id}")).to_not be_checked
+    expect(find("#user_radio_#{users[1].id}")).to be_checked
+    expect(find("#user_radio_#{users[2].id}")).to_not be_checked
+  end
+
+
+
+
+end
```
bobbytables: Formatting :)

spec/features/user_views_list_of_users_and_owners_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    click_button "Save"
+
+    visit dashboard_path(as: users[0])
+    click_link t("dashboards.show.view_all_users")
+
+    expect(find("#user_radio_#{users[0].id}")).to_not be_checked
+    expect(find("#user_radio_#{users[1].id}")).to be_checked
+    expect(find("#user_radio_#{users[2].id}")).to_not be_checked
+  end
```
bobbytables: This scenario is great!

bobbytables:
Minus the one comment this looks great :)

app/controllers/users_controller.rb
ellisbenjamin: Thanks for the catch.  Implemented as:

```
  def update
    installation = current_user.installation
    if user = installation.users.find(params[:user_radio])
      installation.update(owner_id: user.id)
      redirect_to users_path
    else
      redirect_to users_path
    end
  end
```

spec/features/user_views_list_of_users_and_owners_spec.rb
ellisbenjamin: woops.  fixed.

app/views/users/index.html.erb
ellisbenjamin: Moved that td tag from line 15 to 14

app/controllers/users_controller.rb
```diff
@@ -1,5 +1,17 @@
...
+    @installation_id = current_user.installation.id
     @users = current_user.installation.users
+    @owner_id = current_user.installation.owner.id
+  end
+
+  def update
+    installation = current_user.installation
+    if user = installation.users.find(params[:user_radio])
+      installation.update(owner_id: user.id)
+      redirect_to users_path
```
cored: @ellisbenjamin Looks like you always redirecting, why not move that line into the button and remove the else? Also you could push this logic into the model and make it like. 
Also looks like that what you doing with the data is, an update of the owner, what do you think about having a method call like: `installation.update_owner(params[:user_radio])` ?

cored:
:+1:

app/controllers/users_controller.rb
bobbytables: `update_owner` is a good idea.

ellisbenjamin: thanks @cored!

I gave the Installation model an update_owner method and updated the redirect logic.  We are always redirecting back to the users_path so I updated that little bit of logic too.  I'm not sure that I understand moving the redirect to the button

```
  def update_owner(owner_id)
    update_attribute(:owner_id, owner_id)
  end
```

PR #314, opened 2016-06-01
Bellis core 104

ellisbenjamin:
https://namely.atlassian.net/browse/CORE-104

As Namely Connect, I need to capture data about clients that have disconnected any Connector so that I understand more about the usage.

Conditions of Satisfaction:
Namely Connect will capture the following information if a client disconnects any of the supported connectors (Greenhouse, iCims, Jobvite, NetSuite)
* Client Name
* Connector Name
* Date/time stamp that connector was disconnected


spec/models/greenhouse/connection_spec.rb
```diff
@@ -29,6 +29,20 @@
...
 
+  describe '#log_destroy' do
+    it 'creates a log of disconnection' do
+      greenhouse_connection = create :greenhouse_connection, :connected
+      expect(greenhouse_connection.installation_id).to_not be_nil
+
+      installation_id = greenhouse_connection.installation_id
+      greenhouse_connection.destroy
+
+      expect(Disconnection.last().installation_id).to be(installation_id)
```
bobbytables: Style comment: No need for the `last()`, can be just `last.installation_id`

spec/models/icims/connection_spec.rb
```diff
@@ -25,4 +25,17 @@
...
+
+  describe '#log_destroy' do
+    it 'creates a log of disconnection' do
+      icims_connection = create :icims_connection, :connected
+      expect(icims_connection.installation_id).to_not be_nil
+
+      installation_id = icims_connection.installation_id
+      icims_connection.destroy
+
+      expect(Disconnection.last().installation_id).to be(installation_id)
```
bobbytables: Ditto.

app/models/greenhouse/connection.rb
```diff
@@ -60,5 +61,12 @@ def integration_id
     def set_secret_key
       self.secret_key = SecureRandom.hex(20)
     end
+
+    def log_destroy
+      d = Disconnection.new()
```
bobbytables: This can be a bit shorter:

```ruby
Disconnection.create(client_name: integration_id, installation: installation)
```

db/migrate/20160531193238_add_disconnection_logging.rb
```diff
@@ -0,0 +1,9 @@
+class AddDisconnectionLogging < ActiveRecord::Migration
+  def change
+      create_table :disconnections do |t|
+        t.string :client_name
```
bobbytables: I'd recommend renaming this to `integration_name` because client is confusing since this logs "net_suite". Client name would be something like "benellis-sandbox" which is not the case here.

app/models/net_suite/connection.rb
```diff
@@ -80,6 +82,13 @@ def client
     NetSuite::Client.from_env.authorize(authorization)
   end
 
+  def log_destroy
+    d = Disconnection.new()
+    d.client_name = integration_id
+    d.installation_id = installation_id
+    d.save
+  end
```
bobbytables: This `log_destroy` method could probably be put into a ruby module and included in these connection classes to reduce the amount of duplication.

spec/models/greenhouse/connection_spec.rb
ellisbenjamin: fixed

spec/models/icims/connection_spec.rb
ellisbenjamin: fixed

app/models/greenhouse/connection.rb
ellisbenjamin: fixed in each model in which this appears.  will read this mixins doc

db/migrate/20160531193238_add_disconnection_logging.rb
ellisbenjamin: agreed.

app/models/net_suite/connection.rb
ellisbenjamin: good advice!  thanks.  done.  this is in the helpers directory.

app/helpers/log_destroy.rb
```diff
@@ -0,0 +1,5 @@
+module Log_destroy
```
bobbytables: this should be `LogDestroy`

app/helpers/log_destroy.rb
```diff
@@ -0,0 +1,5 @@
...
+  def log_destroy
+      Disconnection.create(integration_name: integration_id, installation_id: installation_id)
```
bobbytables: tabbing size is too big?

ellisbenjamin: word, fixed.

app/helpers/log_destroy.rb
ellisbenjamin: fixed!

bobbytables:
👍  Nice job!

PR #318, opened 2016-06-09
Bellis core 107

ellisbenjamin:
https://namely.atlassian.net/browse/CORE-107

As a Namely Client Admin logging in with basic http auth, I want to see how many total profiles were imported and created via the Greenhouse Connector.
Conditions of Satisfaction:
* Greenhouse Connector is logging number of profiles imported
* Greenhouse Connector is logging number of profiles created via Namely API
* Able to report on these stats

I've created a generic model called ProfileImport that should work with greenhouse, icims, etc.  This PR lets namely connect log all successful greenhouse candidate imports, but there is not a view for setup yet.


db/migrate/20160602154525_create_profile_imports.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.integer :installation_id
+      t.integer :applicant_id
+      t.string :secret_key
+      t.string :first_name
+      t.string :last_name
+      t.string :email, :unique => true
+      t.boolean :successful
+    end
+  end
+end
```
johnnymo87: Does this need any null constraints, defaults, or indexes?

app/services/greenhouse/candidate_import_assistant.rb
```diff
@@ -33,6 +33,16 @@ def import_candidate
         @import_candidate ||= context.namely_importer.single_import(
           greenhouse_payload
         )
+        if @import_candidate.success?
+          ProfileImport.create(
+            installation_id: connection.installation_id,
+            integration_id: connection.integration_id,
+            applicant_id: person_id,
+            first_name: params[:payload][:application][:candidate][:first_name],
+            last_name: params[:payload][:application][:candidate][:last_name]
```
johnnymo87: Are any of these intermediate hashes ever `nil`?

db/migrate/20160602154525_create_profile_imports.rb
ellisbenjamin: Yep, it will.  I'm planning on using this table to store imports from Jobvite and Icims connections as well, (issues CORE-106 and CORE-108), which may provide slightly different data.  I'm fine with holding off on this PR until the really-finalized version of the schema is figured out, or using this as a first-pass and finalizing it when the last of the other related stories is completed.

app/services/greenhouse/candidate_import_assistant.rb
ellisbenjamin: They're present in the greenhouse payload fixtures that were initially setup for testing.  I don't think they will ever be nil but happy to add checks if you think it's a good idea.

PR #319, opened 2016-06-10
Bellis core 106

ellisbenjamin:
https://namely.atlassian.net/browse/CORE-106

As a Namely Client Admin, I want to see how many total profiles were imported and created via the iCims Connector.
Conditions of Satisfaction:
* iCims Connector is logging number of profiles imported
* iCims Connector is logging number of profiles created via Namely API
* Able to report on these stats

Okay.  The ProfileImport model from CORE-107 will handle holding the data and the import_candidate function from
module Icims
  class CandidateImportAssistant
takes care of populating the model when a candidate is succesfully imported.


therealplato:
+1

PR #320, opened 2016-06-15
Bellis core 320

ellisbenjamin:
As a Namely Connect Admin
I want to see information about a company
So that I can make informed decisions when supporting them

--

Then I see a page with the installation's enabled connections

https://namely.atlassian.net/browse/CORE-320

all cred to nick.


app/controllers/admin/installations_controller.rb
```diff
@@ -20,7 +20,24 @@ def show
...
+      @installation = Installation.find(params[:id])
+      @connections = get_installation_connections(params[:id])
+    end
+  end
+
+  private
+
+  def get_installation_connections(i_id)
+    connections = []
+    INTEGRATIONS.each do |k,v|
```
johnnymo87: Would `map` work here?

app/controllers/admin/installations_controller.rb
```diff
@@ -20,7 +20,24 @@ def show
           @installations << installation
         end
       end
+      render :index
+
+    else
+      @installation = Installation.find(params[:id])
```
cored: You could remove a little bit of duplication by having a method that do this lookup. 
```
def installation
@installation ||= Installation.find(params[:id])
end
```

config/routes.rb
```diff
@@ -39,7 +39,6 @@
   )
 
   namespace :admin do
-    resources :installations
+    resources :installations, only: [:index, :show]
```
cored: Good call

cored:
:+1:

app/controllers/admin/installations_controller.rb
```diff
@@ -20,7 +20,24 @@ def show
...
+
+  private
+
+  def get_installation_connections(i_id)
+    connections = []
+    INTEGRATIONS.each do |k,v|
+      connection = Installation.find(i_id).connection_to(k)
+      if connection.connected?
+        connections << connection
+      end
```
bobbytables: We can do `each_with_object`, like so:

```ruby
INTEGRATIONS.each_with_object([]) do |(k, v), connections|
  connection = Installation.find(i_id).connection_to(k)
  if connection.connected?
    connections << connection
  end
end
```

nnarh:
added @bobbytables and @cored suggestions with @ellisbenjamin

bobbytables:
👍

PR #324, opened 2016-06-15
Admin can view list of users on installation

nnarh:
As a Namely Connect Admin
I want to see everyone that has logged into a companies installation
So that I can make informed decisions when supporting them
–
Then I see a page with the installation's users

https://namely.atlassian.net/browse/CORE-323


app/controllers/admin/installations_controller.rb
```diff
@@ -38,6 +39,10 @@ def get_installation_connections(i_id)
     end
   end
 
+  def get_installation_users(i_id)
+    @users = User.where(:installation_id => i_id)
```
bobbytables: Can't we do `installation.users` here since we have the installation?

app/views/admin/installations/show.html.erb
```diff
@@ -14,4 +14,19 @@
         </tr>
     <% end %>
   </tbody>
+
+  <thead>
+    <tr>
+      <th>Users</th>
```
bobbytables: Probably just `Name` right?

Also, can we include the email and created at date please?

rwalters:
Just a quick Jira tip: If you include the "slug" of the Jira ticket (in this case, `CORE-323`) in the title when you create the PR, it will link in the Jira ticket in a "Development" section. Also, if you include the slug in a commit message, a link to that commit will automatically appear in the Jira ticket.

therealplato:
thx @rwalters

nnarh:
awesome! @rwalters

spec/features/admin/admin_visits_installations_page_spec.rb
```diff
@@ -41,4 +41,21 @@ def http_auth
     expect(page).to have_content("Greenhouse")
     expect(page).to_not have_content("Net suite")
   end
+
+  scenario 'Admin can see number of connectioned users per installation' do
```
bobbytables: typo: "connectioned"

app/views/admin/installations/show.html.erb
```diff
@@ -11,7 +13,34 @@
           <td>
             <%= "#{connection.integration_id}".humanize %>
           </td>
+          <td>
+          </td>
+          <td>
+          </td>
         </tr>
     <% end %>
   </tbody>
```
bobbytables: I really think these should be separate tables in the HTML.

bobbytables: Ideally you'd have a `<h3>` or something describing the tables and to separate them nicely.

PR #325, opened 2016-06-17
bellis-hotfix-2016-06-17

ellisbenjamin:
hotfix, if salary comes in from greenhouse with no value field, we put salary at zero rather than throwing a server error.


therealplato:
+1, I wonder how unknown salaries are represented in hcm and payroll

johnnymo87:
+1

PR #330, opened 2016-06-20
CORE-324

nnarh:
As a Namely Connect Admin
I want to change an installations owner
So that when people leave a company or permissions change their installation can continue to work
--

Then I see a page with the installation's users
And I can change the ownership of the installation to a new user

https://namely.atlassian.net/browse/CORE-324


app/assets/stylesheets/_tables.css.scss
```diff
@@ -48,3 +48,23 @@ table {
...
+
+.users {
+  table-layout: fixed;
+  width: 100%;
+  white-space: nowrap;
+}
+
+.users td {
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
bobbytables: 🔥

bobbytables:
:shipit: 🚢

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Admin::UsersController < Admin::BaseController
+  def update
+    if user = User.find(params[:user_radio])
+      user.installation.update_owner(user.id)
```
therealplato: can a user be part of multiple installations?
if so, this is probably not a correct installation lookup

app/assets/stylesheets/_tables.css.scss
```diff
@@ -48,3 +48,23 @@ table {
...
 }
+
+.users {
+  table-layout: fixed;
+  width: 100%;
+  white-space: nowrap;
+}
+
+.users td {
+  white-space: nowrap;
```
poopoothegorilla: if you apply this on `.users` does it no cascade to `td` elements? fyi It has been a while since i worked with css... also any reason you chose not to nest here?

app/views/admin/installations/show.html.erb
```diff
@@ -1,4 +1,8 @@
 <h1><%= @installation.subdomain %></h1>
+
 <%= render "admin/installations/tables/active_connections" %>
 
-<%= render "admin/installations/tables/active_users" %>
+<%= form_tag("/admin/users", method: :patch)  do-%>
```
therealplato: if this is where a user chooses the new owner, you need some `<p>`, `<span>`, `<label>` etc describing what action the user is about to take. Unless of course that's included in a parent template

app/controllers/admin/users_controller.rb
bobbytables: The hierarchy is `Installation <- User`. Users can only have one installation.

app/views/users/_user_table.html.erb
```diff
@@ -0,0 +1,19 @@
+<table class="users">
+  <tr>
+    <h3 class="active-users-header">Connected Users</h3>
+    <th class="owner-header">Owner?</th>
```
therealplato: ^ got it, nvm last comment

app/assets/stylesheets/_tables.css.scss
poopoothegorilla: lol

app/views/users/_user_table.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <th class="first-name-header">First Name</th>
+    <th class="last-name-header">Last Name</th>
+    <th class="email-header">Email</th>
+    <th class="created-at-header">Created At</th>
+  </tr>
+  <% @users.each do |user| %>
+    <tr>
+      <td><%= radio_button_tag("user_radio", user.id, user.id == owner_id) %></td>
```
therealplato: is there any way you can end up in this template with a mistaken or blank owner_id? if so I suspect the value of the form key `user_radio` will be `""` which might have unexpected behavior when posted

spec/features/admin/admin_visits_installations_page_spec.rb
```diff
@@ -53,9 +54,61 @@ def http_auth
...
+
+    click_button "Save"
+
+    visit admin_installations_path
+    click_link installation.subdomain
+
+    expect(find("#user_radio_#{users[0].id}")).to_not be_checked
+    expect(find("#user_radio_#{users[1].id}")).to be_checked
+    expect(find("#user_radio_#{users[2].id}")).to_not be_checked
+  end
```
therealplato: out of curiosity, is every installation admin also the installation owner?

app/assets/stylesheets/_tables.css.scss
therealplato: ha was this a problem before?

poopoothegorilla: 🚒

poopoothegorilla: 🚑

poopoothegorilla:
+1

app/assets/stylesheets/_tables.css.scss
ellisbenjamin: 😐

app/views/users/_user_table.html.erb
ellisbenjamin: The installation model is setup such that if an owner_id isn't explicitly set on an installation, the owner is identified as the first user in the database belonging to the installation.  Since you need to belong to a namely installation in order to login to namely connect, this case should never happen.

PR #331, opened 2016-06-22
bellis-core-108

ellisbenjamin:
https://namely.atlassian.net/browse/CORE-108

Conditions of Satisfaction:
- Jobvite Connector is logging number of profiles imported
- Jobvite Connector is logging number of profiles created via Namely API
- Able to report on these stats

main change is just to /models/jobvite/connection.rb 

for each profile it takes from jobvite we add a record to the ProfileImport model
```
    def import_profile(profile)
      ProfileImport.create(
        installation_id: installation_id,
        integration_id: integration_id,
        applicant_id: profile.profile_id,
        first_name: profile.first_name,
        last_name: profile.last_name,
        email: profile.email
      )
    end
```

This PR includes the already-approved code from 106, 107, and 109.


app/models/jobvite/client.rb
```diff
@@ -40,7 +40,9 @@ def initialize(connection:, start: 1)
 
       def candidates
         process_errors
-        json_response.fetch("candidates", {}).map { |hash| Candidate.new(hash) }
+        profiles = json_response.fetch("candidates", {}).map { |hash| Candidate.new(hash) }
+
+        return profiles
```
bobbytables: Why this change?

bobbytables: It's the same with the old diff I think.

app/models/jobvite/connection.rb
```diff
@@ -57,9 +57,27 @@ def required_namely_field
...
+      profiles = import_results.map do |result|
         Result.new(**result)
       end
+
+      profiles.each do |profile|
+        if profile.success?
+          import_profile(profile)
+        end
+      end
+      return profiles
```
bobbytables: Lets not use explicit returns unless we need to.

app/models/jobvite/client.rb
ellisbenjamin: Yep, you're right.  That was an artifact from a failed attempt at implementing this feature a different way.

app/services/greenhouse/candidate_import_assistant.rb
```diff
@@ -33,6 +33,18 @@ def import_candidate
...
         )
+        if @import_candidate.success?
+          emails = params[:payload][:application][:candidate][:email_addresses]
+          ProfileImport.create(
+            installation_id: connection.installation_id,
+            integration_id: connection.integration_id,
+            applicant_id: params[:payload][:application][:candidate][:id],
+            first_name: params[:payload][:application][:candidate][:first_name],
+            last_name: params[:payload][:application][:candidate][:last_name],
+            email: emails.find {|email| email["type"] == "work"}["value"]
```
bobbytables: This kinda scares me because if the email isn't found at all we're gonna be calling `nil["value"]`

app/views/admin/installations/index.html.erb
```diff
@@ -4,6 +4,7 @@
 
 <p>
   Filter by
+  <%= link_to "All", admin_installation_path(id: "") %>,
```
bobbytables: Is the `id: ""` needed?

db/migrate/20160602154525_create_profile_imports.rb
```diff
@@ -0,0 +1,15 @@
+class CreateProfileImports < ActiveRecord::Migration
+  def change
+    create_table :profile_imports do |t|
+      t.string :integration_id
+      t.integer :installation_id
```
bobbytables: These 2 should have the option `null: false` set.

app/views/admin/installations/index.html.erb
ellisbenjamin: Ah ha.. Changed to  admin_installations_path for all these links and now `id: ""` is not needed.

db/migrate/20160602154525_create_profile_imports.rb
ellisbenjamin: 👍

app/services/greenhouse/candidate_import_assistant.rb
ellisbenjamin: @bobbytables would a string of fetch('key',{}).fetch('key', 'no@email.com') or similar be an acceptable fix?

ellisbenjamin: @bobbytables         if @import_candidate.success? will return false if the email address is not present as described on line 37, so we won't reach line 37 in the event that an email is not found.

PR #336, opened 2016-06-27
INT-3 - Namely Connect - Capture Metrics on When Last Used

ellisbenjamin:
As Namely Connect, I want to capture when the last time a Connector was used and by whom (client) so that it will help me resolve any issues.

Conditions of Satisfaction:
Namely Connect logs the following information for every time a connector is used:
* Name of Connector
* Client name
* Datetime stamp used 
* Status of integration (Success, Fail, NA)

https://namely.atlassian.net/browse/INT-3


app/models/jobvite/connection.rb
```diff
@@ -75,7 +77,20 @@ def import_profile(profile)
...
+    end
+
+    def import_profile_failed(profile)
+      ProfileImport.create(
+        installation_id: installation_id,
+        integration_id: integration_id,
+        applicant_id: profile.profile_id || nil,
+        first_name: profile.first_name || nil,
+        last_name: profile.last_name || nil,
+        email: profile.email || nil,
```
stephwag: You don't need `|| nil` here if you just want to store nil when it's nil

app/models/jobvite/connection.rb
```diff
@@ -75,7 +77,20 @@ def import_profile(profile)
         applicant_id: profile.profile_id,
         first_name: profile.first_name,
         last_name: profile.last_name,
-        email: profile.email
+        email: profile.email,
+        successful: 1
```
stephwag: Since this is the same as `import_profile_failed` with only a success flag, can't you just do something like `success: profile.success?`

app/jobs/net_suite_export_job.rb
```diff
@@ -52,7 +52,18 @@ def perform(operation, sync_summary_id, net_suite_connection_id,
...
+      )
+    else
+      ProfileImport.create(
+        installation_id: connection.installation_id,
+        integration_id: connection.integration_id,
+        applicant_id: netsuite_id || nil,
+        first_name: netsuite_employee.fetch('firstName',nil),
+        last_name: netsuite_employee.fetch('lastName', nil),
+        email: netsuite_employee.fetch('email', nil),
+        successful: false
```
stephwag: Same as other comment

app/services/icims/candidate_import_assistant.rb
```diff
@@ -27,7 +27,18 @@ def import_candidate
...
+        )
+      else
+        ProfileImport.create(
+          email: candidate.email || nil,
+          installation_id: connection.installation_id,
+          integration_id: connection.integration_id,
+          applicant_id: person_id || nil,
+          first_name: candidate.firstname || nil,
+          last_name: candidate.lastname || nil,
+          successful: false
```
stephwag: Same as other comment, and you don't need the `|| nil`

spec/jobs/net_suite_export_job_spec.rb
```diff
@@ -62,6 +62,7 @@
         expect(profile.first_name).to eq("NAME")
         expect(profile.last_name).to eq("LAST")
         expect(profile.email).to eq("e@mail.com")
+        expect(profile.successful).to eq(true)
```
stephwag: Style suggestion, you can also use `be_true` instead of `eq(true)`

app/models/jobvite/connection.rb
ellisbenjamin: Yes, that is much nicer.  Thanks 👍

spec/jobs/net_suite_export_job_spec.rb
ellisbenjamin: I gave this a whirl and it fails -- did a little research and found this:

```Yep. be_true was deprecated in 2.99 and has been removed in 3.0. Use be_truthy (if you want the same semantics as before) or be true (or eq true, or whatever) if you want exact equality semantics.```

our other tests use .eq(true) so I think I'll stick with it unless anyone objects.

PR #337, opened 2016-06-29
Form errors now orange.

ellisbenjamin:
Errors now orange because they are otherwise very easy to ignore.


bobbytables:
:shipit:

PR #347, opened 2016-07-14
nnarh-int-52

nnarh:
As a user using Namely Connect, I need to be able to view all the setup instructions and/or FAQ for each of the connects that are supported.

https://namely.atlassian.net/browse/INT-52


ellisbenjamin:
👍

